PR #1, opened 2020-03-05
Fix TMPDIR backup

acowley:
This was clearing TMPDIR since it was first not setting the
`tmp_backup` local variable, and then, since `tmp_backup` was null, it
was unsetting TMPDIR.

### DIFF
--- a/direnvrc
+++ b/direnvrc
@@ -43,7 +43,7 @@ use_nix() {
     log_status using cached derivation
   fi
   local term_backup=$TERM path_backup=$PATH
-  if [[ -z ${TMPDIR+x} ]]; then
+  if [[ ! -z ${TMPDIR+x} ]]; then
     local tmp_backup=$TMPDIR
   fi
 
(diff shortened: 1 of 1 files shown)


direnvrc
```diff
@@ -43,15 +43,15 @@ use_nix() {
...
   local term_backup=$TERM path_backup=$PATH
-  if [[ -z ${TMPDIR+x} ]]; then
+  if [[ ! -z ${TMPDIR+x} ]]; then
     local tmp_backup=$TMPDIR
   fi
 
   log_status eval "$cache"
   read -r cache_content < "$cache"
   eval "$cache_content"
   export PATH=$PATH:$path_backup TERM=$term_backup TMPDIR=$tmp_backup
```
Mic92:
```suggestion
  export PATH=$PATH:$path_backup TERM=$term_backup 
```

direnvrc
```diff
@@ -43,15 +43,15 @@ use_nix() {
...
-  if [ -z ${tmp_backup+x} ]; then
+  if [ ! -z ${tmp_backup+x} ]; then
```
Mic92:
```suggestion
  if [[ ! -z ${tmp_backup+x} ]]; then
```

PR #2, opened 2020-03-05
Update README.md with nix.conf changes needed for persistent derivations

matt-snider:
It appears that the derivations produced by nix-direnv are removed when running nix's garbage collection. Troubleshooting this issue led me to [a git issue][1] in the NixOS/nix repo. Enabling both `keep-derivations` and `keep-outputs` worked for me as recommended, and now nix-direnv is working.

[1]: https://github.com/NixOS/nix/issues/2208


Mic92:
Indeed. I added `keep-outputs` some time ago and forgot about it.

Issue #11, opened 2020-04-09
Optionally move project-wise .direnv to a centralized location


yipengsun:
Hi,

Currently, this `use_nix` function will create a `.direnv` folder along with the `.envrc` file. The problem is: I have some of my project stored in Dropbox, and this has 2 problems:

1. Dropbox will mess up symblinks
2. The `cache-` and `drv` files doesn't seem to be portable

Is it possible to add an additional environmental variable to store all `.direnv` in a centralized location?

Mic92:
cc @zimbatm ?

Mic92:
Would it work for you to use the exclude option? 

```
$ dropbox exclude add **.direnv
```

zimbatm:
it's actually already possible but not documented enough: https://github.com/direnv/direnv/commit/faf5fc1c93453cafe1ab8bc8c7c3e8444599fc07

yipengsun:
> Would it work for you to use the exclude option?
> 
> ```
> $ dropbox exclude add **.direnv
> ```

`dropbox` (or `dropbox-cli`) doesn't seem to be available on macOS (There's a `dbxcli` package, but that one doesn't have "excluding" capability). I'll try the `direnv_layout_dir` option.

yipengsun:
I tried to add these lines to `$HOME/direnv/direnvrc`:
```
: ${XDG_CACHE_HOME:=$HOME/.cache}
pwd_hash=$(echo -n $PWD | shasum | cut -d ' ' -f 1)
direnv_layout_dir=$XDG_CACHE_HOME/direnv/layouts/$pwd_hash
```

And I created the `direnv` folder in `$HOME/.cache`. Still, no hashed folder is created. Actually, I think now `direnv` never caches anything.

yipengsun:
BTW the `dropbox` tool does not handle wildcards if this [answer](https://superuser.com/questions/469776/how-to-exclude-files-not-folders-from-dropbox-sync) is true.

Mic92:
Well. `nix-direnv` does not uses `direnv_layout_dir` yet.

Mic92:
> BTW the `dropbox` tool does not handle wildcards if this [answer](https://superuser.com/questions/469776/how-to-exclude-files-not-folders-from-dropbox-sync) is true.

No but your shell should.

yipengsun:
[...]

PR #14, opened 2020-04-11
Fix persistent derivation when not using default direnv-layout-dir

yipengsun:
As mentioned in a later post in #11, current `master` breaks the persistent derivation (prevent-garbage-collection) feature when using a non-default `direnv_layout_dir`.

Here's my naive take on fixing that. I've tried this "fix" both with and without `direnv_layout_dir` explicitly set, and it appears to be working.

The main problem with this fix is that now `realpath` is needed (part of the `coreutils` derivation in Nix).

### DIFF
--- a/direnvrc
+++ b/direnvrc
@@ -84,7 +84,7 @@ use_nix() {
     local escaped_pwd=${stripped_pwd//-/--}
     local escaped_pwd=${escaped_pwd//\//-}
     ln -fs "$drv" "$drv_link"
-    ln -fs "$PWD/$drv_link" "/nix/var/nix/gcroots/per-user/$LOGNAME/$escaped_pwd"
+    ln -fs "$(realpath -s $drv_link)" "/nix/var/nix/gcroots/per-user/$LOGNAME/$escaped_pwd"
     log_status renewed cache and derivation link
   fi
 


yipengsun:
I spotted the following error in the CI:
```
ln: failed to create symbolic link '/nix/var/nix/gcroots/per-user//tmp-tmpt2usctww-testenv': Permission denied
```

But I think this is due to some CI limitation.

Mic92:
mhm maybe no `$USER` is set?

direnvrc
```diff
@@ -84,7 +84,7 @@ use_nix() {
     local escaped_pwd=${stripped_pwd//-/--}
     local escaped_pwd=${escaped_pwd//\//-}
     ln -fs "$drv" "$drv_link"
-    ln -fs "$PWD/$drv_link" "/nix/var/nix/gcroots/per-user/$LOGNAME/$escaped_pwd"
+    ln -fs "$(realpath -s $drv_link)" "/nix/var/nix/gcroots/per-user/$LOGNAME/$escaped_pwd"
```
Mic92, line 87: Actually `$PWD` should be now dropped in both cases.

Mic92:
Fixed in https://github.com/nix-community/nix-direnv/pull/15

PR #21, opened 2020-05-25
readme: update home-manager instructions

Valodim:
Update readme with a more home-managery way of installing nix-direnv in home-manager.


Mic92:
Thanks!

PR #22, opened 2020-06-07
Updated installation guide for home-manager

pca006132:
Home-manager now supports the option [programs.direnv.enableNixDirenvIntegration](https://rycee.gitlab.io/home-manager/options.html#opt-programs.direnv.enableNixDirenvIntegration).
User does not need to set the `stdlib` option now.


Mic92:
Thanks!

PR #29, opened 2020-07-17
Mark NIX_SHEL as impure

### LINKED ISSUE #28: IN_NIX_SHELL should be pruned
I noticed that in my env `IN_NIX_SHELL=pure` after using this plugin; but that seems incorrect since the premise of this plugin is to merge the NIX_SHELL into your other shell; making it impure.

Might be worthwhile to purge this variable from the environment dump.
It was very noticeable because my ZSH theme was picking up that environment variable and displaying it.

image

fzakaria:
direnv uses a pure nix-shell to figure out environment variables
which it then loads into your normal shell; making the final shell
being used impure.

Set the environment variable correctly to identify it as so.

Fix #28


fzakaria:
Ran it locally:

```bash
❯ env | grep NIX_SHELL
IN_NIX_SHELL=impure
```

image

fzakaria:
Off topic: but I am using [powerlevel10k](https://github.com/romkatv/powerlevel10k) as my theme.
I need to figure out why it's using _IN_NIX_SHELL_ in the prompt as opposed to the `name`

```
❯ env | grep name
name=nix-shell
```

direnvrc
```diff
@@ -60,6 +60,11 @@ use_nix() {
     unset TMPDIR
   fi
 
+  # `nix-shell --pure` sets IN_NIX_SHELL to pure which is
+  # misleading since we are in an impure shell now
+  # https://github.com/nix-community/nix-direnv/issues/28
+  export IN_NIX_SHELL=impure
```
Mic92, line 66:
```suggestion
  # `nix-shell --pure` sets IN_NIX_SHELL to pure which is
  # misleading since we are in an impure shell now
  export IN_NIX_SHELL=impure
```

fzakaria:
yay contribution!

Mic92:
You are welcome.

PR #31, opened 2020-08-06
add use_flake

Mic92:


direnvrc
```diff
@@ -1,5 +1,52 @@
...
+
+  # misleading since we are in an impure shell now
+  export IN_NIX_SHELL=impure
+}
+
+use_flake() {
+  watch_file flake.nix
+  watch_file flake.lock
+
+  _nix_import_env "$(nix print-dev-env --profile "$(direnv_layout_dir)/flake-profile")"
```
Mic92, line 47: @zimbatm what do you think about this here?

```suggestion
  _nix_import_env "$(nix --no-net print-dev-env --profile "$(direnv_layout_dir)/flake-profile")"
```

I often encountered nix to forget about downloaded flakes. An alternative would be gc roots for all inputs.

Mic92:
The ci needs to be upgraded to `nixUnstable`.

direnvrc
zimbatm: `--no-net` is new to me. I did a quick try on one of my repos and it started re-compiling perl and gettext. Without `--no-net` it was able to fetch those results from the cache.

I think the issue you are mentioning is legit but needs to be solved differently. The `--profile` only captures the build outputs, and the inputs can get garbage-collected. To solve that, one way would be to somehow capture all the inputs into the build output. Or create a separate buildEnv profile with the inputs.

Mic92: Actually I did a test and it also did not protect the outputs from beeing garbage collected.
I added a gcroot for the profile.

Mic92: I think the the directory specified in `--profile` must point to `/nix/var/nix/gcroots/per-user/$USER` to be effective.

zimbatm:
> The ci needs to be upgraded to `nixUnstable`.

see https://github.com/numtide/nix-flakes-installer#github-actions

direnvrc
```diff
@@ -1,5 +1,122 @@
 # shellcheck shell=bash
 
+_nix_import_env() {
+  local env=$1
+
+  local term_backup=$TERM path_backup=$PATH
+  if [[ -n ${TMPDIR+x} ]]; then
```
zimbatm, line 7: The `+x` trick is not needed in bash anymore, you can use `if [[ -z $TMPDIR ]]; then`. In POSIX sh you would also have to escape the `"$TMPDIR"`, which is also not needed when using the `[[ ]]` test syntax (but only on the RHS).

direnvrc
```diff
@@ -1,5 +1,122 @@
...
+    local tmp_backup=$TMPDIR
+  fi
+  local impure_ssl_cert_file=${SSL_CERT_FILE:-__UNSET__}
```
zimbatm, line 10: this could be simplified with `local impure_ssl_cert_file=$SSL_CERT_FILE` and then use the `-z` test below

zimbatm: or `local impure_ssl_cert_file=${SSL_CERT_FILE:-}` if `set -u` is being used actually

direnvrc
```diff
@@ -1,5 +1,122 @@
...
+
+  eval "$env"
+
+  # `nix-shell --pure` sets invalid ssl certificate paths
+  if [[ "${SSL_CERT_FILE:-}" = /no-cert-file.crt ]]; then
+    if [[ $impure_ssl_cert_file == __UNSET__ ]]; then
+      unset SSL_CERT_FILE
+    else
+      export SSL_CERT_FILE=$impure_nix_ssl_cert_file
+    fi
```
zimbatm, line 21: some utility function like that could simplify the code:
```sh
maybe_export() {
  local key=$1
  local value=$2
  if [[ -z $value ]]; then
    unset "$key"
  else
    export "$key=$value"
  fi
}
```
}

direnvrc
```diff
@@ -1,5 +1,122 @@
...
+
+  local profile="$(direnv_layout_dir)/flake-profile"
+  local profile_rc="${profile}.rc"
+
+  if [[ ! -e "$profile" ]] || \
+    [[ ! -e "$profile_rc" ]] || \
+    [[ "$HOME/.direnvrc" -nt "$profile_rc" ]] || \
+    [[ .envrc -nt "$profile_rc" ]] || \
+    [[ flake.nix -nt "$profile_rc" ]] || \
+    [[ flake.lock -nt "$profile_rc" ]];
```
zimbatm, line 66: the `[[ ]]` test keyword also supports `||` inside it

direnvrc
Mic92: How would distinguish an empty string from an unset one in this case?

direnvrc
```diff
@@ -1,5 +1,122 @@
...
+
+  local old_nix_build_top=${NIX_BUILD_TOP:-__UNSET__}
+  local old_tmp=${TMP:-__UNSET__}
+  local old_tmpdir=${TMPDIR:-__UNSET__}
+  local old_temp=${TEMP:-__UNSET__}
+  local old_tempdir=${TEMPDIR:-__UNSET__}
+  eval "$(< "$profile_rc")"
+  # nix print-env-dev will create a temporary directory and use it a TMPDIR,
+  # we cannot rely on this directory beeing not deleted at some point,
+  # hence we are just removing it right away.
```
ncfavier, line 88: I don't follow. We can't a priori rely on `$old_tmpdir` not being deleted either, what's the difference? At least `print-dev-env`'s `TMPDIR` is a subfolder of `/tmp`.

Mic92: This is more about a temporary directory being deleted and we still cache the environment variables pointing to it.

Mic92: `old_tmpdir` might be a value set by a user and hopefully is persistent/sane i.e. TMPDIR=/run/user/$UID or simply TMPDIR=/tmp

PR #58, opened 2020-12-30
direnvrc: Fix reading of nixpkgs version from git checkout

DerGuteMoritz:
Without this patch, the following error occurs when using a git checkout of nixpkgs:

```
/run/current-system/sw/share/nix-direnv/direnvrc:115: .git/refs/heads/nixos-unstable: No such file or directory
```


Mic92:
Thanks!

PR #59, opened 2021-01-02
add NIX_BIN_PREFIX to hardcode nix

Mic92:


cpcloud:
I just bisected my configuration, and somehow this commit causes every line of stdout to be duplicated in any project that uses `direnv`.

I'm using `home-manager` with the `programs.direnv.enableNixDirenvIntegration = true;`.

cpcloud:
Ok, after changing `nixFlakes` to `nix` in `default.nix` on `master`, the double logging goes away. Is this a bug in the nix binaries under `nixFlakes`?

cpcloud:
In the meantime, it's easy enough to override `nixFlakes` with `nix` in an overlay.

Mic92:
This was a bug in nixUnstable at some point. It got fixed however eventually. You must use an old version.

cpcloud:
I'm using `niv` to manage `nixpkgs` and running against `nixos-unstable-small`, so I assume that I have the latest-but-not-master-mostly-reliably-passing versions of everything.

PR #63, opened 2021-01-29
Force use of GNU Grep (on OSX)

edude03:
If you install nix-direnv using `source_url` (or any way that doesn't use the nix derivation) on OSX, issue #3 still exists. So I added a "hack" that might not be so much of a hack since if the user is installing this they likely have nix, and if you're using the nix derivation, this should resolve to the same version of grep anyway.

### DIFF
--- a/direnvrc
+++ b/direnvrc
@@ -121,7 +121,7 @@ use_flake() {
 use_nix() {
   local path layout_dir grep
   path=$("${NIX_BIN_PREFIX}nix-instantiate" --find-file nixpkgs)
-  grep=$("${NIX_BIN_PREFIX}nix-instantiate" -E 'with import <nixpkgs> {}; gnugrep')/bin/grep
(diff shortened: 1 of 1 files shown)


direnvrc
```diff
@@ -119,8 +119,9 @@ use_flake() {
 }
 
 use_nix() {
-  local path layout_dir
+  local path layout_dir grep
   path=$("${NIX_BIN_PREFIX}nix-instantiate" --find-file nixpkgs)
+  grep=$("${NIX_BIN_PREFIX}nix-build" -E 'with import <nixpkgs> {}; gnugrep')/bin/grep
```
Mic92, line 124: I am bit worried about people not having NIX_PATH set with nixpkgs
What do you think about this solution? https://github.com/nix-community/nix-direnv/pull/64

edude03: Tested #64 and it worked for me on OSX (didn't test on linux, but it'll be fine I'm sure) so I'll close this. Thanks!

PR #64, opened 2021-01-30
get rid of gnugrep dependency

Mic92:


edude03:
Tried it and it worked for the issue I encountered in #63 - so 👍 for me.

Mic92:
1.2.2 is released now.

PR #69, opened 2021-02-23
Use impure environment for dump command

jgeerds:
Since `/share` directories of `nativeBuildInputs` are added to `$XDG_DATA_DIRS` (see https://github.com/direnv/direnv/issues/443#issuecomment-777730925), the original `$XDG_DATA_DIRS` variable (i.e. before entering the directory / loading the environment) is completely overwritten by the new one.

This just happens when using `nix-direnv`. The original `use_nix` function by `direnv` does not use `--pure` and works as expected:

```
use_nix() {
  direnv_load nix-shell --show-trace "$@" --run "$(join_args "$direnv" dump)"
  if [[ $# == 0 ]]; then
    watch_file default.nix shell.nix
  fi
}
```

This pull request fixes the issue for me.

### DIFF
--- a/direnvrc
+++ b/direnvrc
@@ -187,7 +187,7 @@ use_nix() {
     dump_cmd="echo _____direnv_____; \"$direnv\" dump bash"
     tmp=$("${NIX_BIN_PREFIX}nix-shell" \
       "${experimental_flags[@]}" \
-      --show-trace --pure "$@" --run "$dump_cmd")
+      --show-trace "$@" --run "$dump_cmd")
     # show original shell hook output
     echo "$tmp" | _nix_extract_direnv >&2 > "$cache"
     update_drv=1


direnvrc
```diff
@@ -187,7 +187,7 @@ use_nix() {
     dump_cmd="echo _____direnv_____; \"$direnv\" dump bash"
     tmp=$("${NIX_BIN_PREFIX}nix-shell" \
       "${experimental_flags[@]}" \
-      --show-trace --pure "$@" --run "$dump_cmd")
+      --show-trace "$@" --run "$dump_cmd")
```
Mic92, line 190: I don't think we have to `--pure` otherwise we cannot cache its output since it also include the users environment.

Mic92: Is there maybe a way to save the old value of XDG_DATA_DIRS and than just append the value from nix-shell?

meck: I just hit this issue, should be possible to handle `XDG_DATA_DIRS`  the same way `PATH` is handled in `_nix_import_env()`?

Mic92:
Solved in https://github.com/nix-community/nix-direnv/pull/71

PR #73, opened 2021-03-11
Avoid pitfall in home-manager config instructions

OlivierNicole:
When I followed the README, direnv did not get hooked into zsh because I had zsh installed at the system level, rather than enabled in my `home-manager` configuration. I suggest you make it explicit that it must be the case for the `home.nix` changes to work.


Mic92:
Thanks!

PR #75, opened 2021-03-19
Dont merge: save XDG_DATA_DIRS in flakes

### LINKED ISSUE #74: XDG_DATA_DIRS not preserved in use_flake
20e2676b1998232bb4d1996e6431b2ee00cee30a removed `_nix_import_env` from `use_flake`, which means the fix in 7ac4290e8247e935fdab5830b2373e41427af237 doesn't apply to `use_flake`, which means `XDG_DATA_DIRS` isn't preserved, which means dynamic shell completion loading fails.

meck:
Should fix #74 but i haven't started to use flakes myself yet, so if someone with a bit more familiar could try it out...

### DIFF
--- a/direnvrc
+++ b/direnvrc
@@ -112,6 +112,7 @@ use_flake() {
   local old_tmpdir=${TMPDIR:-__UNSET__}
   local old_temp=${TEMP:-__UNSET__}
   local old_tempdir=${TEMPDIR:-__UNSET__}
+  local old_xdg_data_dirs=${XDG_DATA_DIRS:-}
   eval "$(< "$profile_rc")"
   # nix print-env-dev will create a temporary directory and use it a TMPDIR,
   # we cannot rely on this directory beeing not deleted at some point,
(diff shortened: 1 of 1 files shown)


ncfavier:
That works! But @Mic92 shouldn't we just reuse `_nix_import_env` in `use_flake`, to avoid code duplication?

Mic92:
I think we could use `_nix_import_env`. I just did not use in the first place since it tries to backup more variables since for nix-shell we use `--pure` evaluation.

direnvrc
```diff
@@ -112,6 +112,7 @@ use_flake() {
   local old_tmpdir=${TMPDIR:-__UNSET__}
   local old_temp=${TEMP:-__UNSET__}
   local old_tempdir=${TEMPDIR:-__UNSET__}
+  local old_xdg_data_dirs=${XDG_DATA_DIRS:-}
```
Mic92, line 115: # TODO: maybe use `_nix_import_env()` here: https://github.com/nix-community/nix-direnv/pull/75#issuecomment-803139886

ncfavier:
Thanks!

PR #76, opened 2021-03-24
Add Missing Code Type Tags

houstdav000:
Enable a little more syntax highlighting in the `README.md` by adding "Nix" to three code blocks missing it, and "bash" to one.

One in-between is intentionally left un-typed due to being a text-based config format rather than a common script or language.


Mic92:
Thanks!

PR #82, opened 2021-05-09
fix shellcheck

Mic92:


Mic92:
Forgot what this was about.

PR #86, opened 2021-05-16
direnvrc: Add `nix_direnv_version`

bb010g:
In the style of direnv's `direnv_version`.
https://github.com/direnv/direnv/blob/729fbecd96f3e827575f19497bc01df33395d679/stdlib.go#L1173-L1179

Can be tested for with `has`, e.g. to decide on using `source_url`.

### DIFF
--- a/direnvrc
+++ b/direnvrc
@@ -1,5 +1,34 @@
 # shellcheck shell=bash
 
+# Usage: nix_direnv_version <version_at_least>
+#
+# Checks that the nix-direnv version is at least as old as <version_at_least>.
+nix_direnv_version() {
+  declare major='1' minor='2' patch='6' # UPDATE(nix-direnv version)
+
+  [[ $1 =~ ^([^+-.]*)(\.?)([^+-.]*)(\.?)([^+-]*)(-?)([^+]*)(\+?)(.*)$ ]]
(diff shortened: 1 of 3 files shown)


direnvrc
```diff
@@ -1,5 +1,34 @@
 # shellcheck shell=bash
 
+# Usage: nix_direnv_version <version_at_least>
```
Mic92, line 3: This does not seem to call nix at all. How does it work?

direnvrc
```diff
@@ -1,5 +1,34 @@
...
+#
+# Checks that the direnv version is at least old as <version_at_least>.
+nix_direnv_version() {
+  declare major=1 minor=2 patch=6
```
Mic92, line 7: Could you bump this version here as well? https://github.com/nix-community/nix-direnv/blob/adfca0d26dc687c1d10a4e88ef32cbc11bcaee47/scripts/create-release.sh#L19

Mic92:
A simple example in the README on how to use this function would be also nice.

bb010g:
https://github.com/nix-community/nix-direnv/blob/db68fb594e9cc26eb67ffa44ce3d3b9e8c3844ba/README.md#L124-L132

Mic92:
Thanks! Merged and released version.

PR #87, opened 2021-05-26
use flake: allow optional flake expression

martinetd:
This allows using flake.nix out of tree, which is very appreciable
if you develop in a big repository as flake evaluation will copy the
whole git archive to store everytime it changes.

To give a more concrete example, I have my dev environments in /etc/nixos which is stable and relatively small, so I use it with package name argument for ech project e.g. `use flake /etc/nixos#linux` in .envrc with:
 http://cgit.notk.org/asmadeus/nixos-config.git/tree/pkgs/linux/default.nix
loaded from http://cgit.notk.org/asmadeus/nixos-config.git/tree/pkgs/default.nix with callPackage
loaded from http://cgit.notk.org/asmadeus/nixos-config.git/tree/flake.nix as output's packages.x86_64-linux

Using an in-repo flake.nix for linux kernel development/testing didn't really work well; I considered an intermediate directory depth but that's also annoying, this is perfect and caching / watch all work well for me.
(I also made sure the no argument case still works)

### DIFF
--- a/README.md
+++ b/README.md
@@ -185,6 +185,12 @@ $ echo "use flake" >> .envrc
 $ direnv allow
 ```
 
+Optionally if you do not want flake.nix to be part of the current directory repo,
+you can specify an arbitrary flake expression as parameter such as:
+```console
+use flake ~/myflakes#project
+```
+
 ## Storing .direnv outside the project directory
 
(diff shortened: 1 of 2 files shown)


Mic92:
Thanks!

PR #91, opened 2021-06-08
flakes: Don't run shellHook on cached dev shells

### LINKED ISSUE #90: shellHook caching inconsistency with and without flakes
In cached shells, the shellHook does not execute with `use nix` but does with `use flake`.

It would be easy to achieve consistency by piping the `nix print-dev-env` output through `sed` to remove the final `eval "$shellHook"`, but I'm not sure this is the right direction. It seems that `nix develop` runs the shellHook even when cached so perhaps it's better not to fight the pattern flakes is currently implementing.

I'm not sure if this is a bug or just a (documentation-worthy?) distinction but it does complicate the creation of environments compatible with or without flakes. For example, I was planning to use an .envrc like:

```
nix flake --help > /dev/null \
  && use flake \
  || use nix
```

but heavy shellHooks will need to be rewritten to check whether their expensive processes have been run already.

ryneeverett:
This makes the behavior of `use flake` consistent with that of `use
nix`.

Resolve #90.

### DIFF
--- a/direnvrc
+++ b/direnvrc
@@ -135,6 +135,7 @@ use_flake() {
 
     log_status renewed cache
   else
+    sed -i "/eval \"\$shellHook\"/d" "$profile_rc"
     log_status using cached dev shell
   fi
 
--- a/tests/test.py
+++ b/tests/test.py
@@ -55,6 +55,7 @@ class IntegrationTest(unittest.TestCase):
         )
         sys.stderr.write(out1.stderr)
         self.assertIn("renewed cache and derivation link", out1.stderr)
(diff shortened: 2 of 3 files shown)


direnvrc
```diff
@@ -135,6 +135,7 @@ use_flake() {
 
     log_status renewed cache
   else
+    sed -i '/eval \"\$shellHook\"/d' "$profile_rc"
```
Mic92, line 138: Can it be not substituted once when it is cached?

Mic92: okay. maybe not worth the complexity.

ryneeverett:
CI should pass now.

PR #93, opened 2021-06-11
 test.py: isolate assertions; DRY data generation 

ryneeverett:
The majority of the test code is just creating the stdout data against
which to make assertions. Increasing the test cases will be more
manageable if the assertions are separate from this data generation.

It is also a generally preferred style to have one assertion per test,
because this way all assertions run even if others fail and each test
only fails for one reason. It also facilitates better test documentation
and adding additional tests without increasing complexity.

### DIFF
--- a/tests/test.py
+++ b/tests/test.py
@@ -30,77 +30,90 @@ def support_flakes() -> bool:
     return proc.stdout == "true"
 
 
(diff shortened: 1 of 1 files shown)


ryneeverett:
Rebased.

PR #99, opened 2021-06-25
chore: add overlay

Hoverbear:
Provide an overlay (and consume it) so that flake users could add a nix module like:

```nix
{
    # ...
    nixosModules = {
        add-overlays = {
            nixpkgs.overlays = [ nix-direnv.overlay ];
            environment.systemPackages = with pkgs; [
                direnv
                nix-direnv
           ];
        };
    };
}
```

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -5,10 +5,16 @@
 
   outputs = { self, nixpkgs, flake-utils }:
     flake-utils.lib.eachDefaultSystem (system: let
-      pkgs = nixpkgs.legacyPackages.${system};
+      pkgs = import nixpkgs {
+        inherit system;
+        overlays = [ self.overlay ];
+      };
     in {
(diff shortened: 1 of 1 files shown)


flake.nix
```diff
@@ -5,10 +5,16 @@
 
   outputs = { self, nixpkgs, flake-utils }:
     flake-utils.lib.eachDefaultSystem (system: let
-      pkgs = nixpkgs.legacyPackages.${system};
+      pkgs = import nixpkgs {
+        inherit system;
+        overlays = [ self.overlay ];
+      };
```
Mic92, line 11: This should be not necessary. It makes evaluation more expensive by re-importing nixpkgs.

Hoverbear: Gosh whoops! Let me fix this.

PR #103, opened 2021-07-29
Add `use_flake_if_supported`

zhaofengli:
This PR adds an opportunistic `use_flake_if_supported` loader which will automatically fall back to `use_nix` if the current Nix version does not support Flakes. In this situation, changes to `flake.nix` and `flake.lock` will still automatically trigger a reload.

This is mostly useful when you provide a `shell.nix` (e.g., though [flake-compat](https://github.com/edolstra/flake-compat)) and nix-direnv is loaded with `source_url` so no changes on the user's end are required.

To test:
```

source_url "https://github.com/zhaofengli/nix-direnv/raw/2121c62181130c2295183421afda26d2ea7549ef/direnvrc" "sha256-zcBGwswqqcm6VY7j+TMzUtqwZF3romx7fIxqkiu3XxI="
```

### DIFF
--- a/direnvrc
+++ b/direnvrc
@@ -86,11 +86,34 @@ _nix_add_gcroot() {
   ln -fsn "$symlink" "/nix/var/nix/gcroots/per-user/$USER/$escaped_pwd"
 }
 
+# -n $(_nix_newer_than /path/to/cache /watched1 /watched2 /watched3)
+_nix_newer_than() {
+  local cache=$1
+  shift
+  local watched=("$@")
+  for f in "${watched[@]}"; do
+    if [[ "$f" -nt "$cache" ]]; then
+      echo "1"
+    fi
+  done
+}
+
+# _nix_add_watched /path/to/file
+declare -a _NIX_DIRENV_WATCHED=()
+_nix_add_watched() {
+  watch_file $1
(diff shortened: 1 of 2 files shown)


zhaofengli:
@Mic92 Gentle poke :eyes:

Mic92:
I see your pain point but I am bit hesitant to merge this because eventually we no longer need this when nix is released ([TM](https://discourse.nixos.org/t/nix-release-schedule-and-roadmap/14204).

I would be ok with linking to your fork for this feature but in long term I prefer to rather not have this:

`source_url` should make installation in different projects easy.

direnvrc
```diff
@@ -86,11 +86,50 @@ _nix_add_gcroot() {
...
+_nix_add_watched() {
+  watch_file $1
+  _NIX_DIRENV_WATCHED+=($1)
+}
+
+_nix_newer_than_watched() {
+  _nix_newer_than $1 "${_NIX_DIRENV_WATCHED[@]}"
+}
+
+use_flake_if_supported() {
```
Mic92, line 112: This function could be also a separate project or gist that one can install on top of `nix-direnv` with `source_url`.

zhaofengli:
Wasn't aware that the 2.4 release is finally on the horizon. That's great news! Let's keep this off-tree for now.

PR #106, opened 2021-08-12
Add a suffix to cache file to allow multiple calls to use_nix

immae:
This change permits to make multiple calls to use_nix, with different
arguments passed to it. It will generate a cache file for each distinct
"argument list" rather than based solely on the nixpkgs version.

### DIFF
--- a/direnvrc
+++ b/direnvrc
@@ -98,7 +98,7 @@ _nix_argsum_suffix() {
       # degrate gracefully both tools are not present
       return
     fi
-    read checksum _ <<< "$out"
+    read -r checksum _ <<< "$out"
     echo "-$checksum"
   fi
 }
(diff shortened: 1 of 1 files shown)


Mic92:
Thanks!

immae:
Thanks for the review @Mic92 ! One comment: you added a `shasum` version, which - I assume - takes the "kind-of-state-of-the-art" hashing method (seems a bit late as on my computer it computes a sha1). This means that whenever the default changes, it will need to recompute everything (and we don’t need to be secure in this case). Is that sensible?

Mic92:
shasum is part of perl while sha1sum comes from coreutils. Both compute sha1 by default. My expectation was that we are more likely to have coreutils than perl.

PR #136, opened 2021-12-21
Allow passing arguments to print-dev-env invocation

### LINKED ISSUE #135: Allow adding or modifying flake parameters
First some background: There is a conflict in behavior with `poetry2nix` and it's editable source behavior and the way that flakes are copied into the nix store that results in the editable source pointing to another spot in /nix/store (as opposed to the local development environment). This can be resolved as follows:

```nix
let 
  PROJECT_BASE_DIR = builtins.getEnv "PROJECT_BASE_DIR";
  editableSources = (if (PROJECT_BASE_DIR != "") then {
    project = "${PROJECT_BASE_DIR}/src";
  } else abort "Must set PROJECT_BASE_DIR");
in (pkgs.poetry2nix.mkPoetryEnv {
  # ...
  editablePackageSources = editableSources;
})
```

The issue here is that this precludes usage with `nix-direnv`'s `use_flake`, as `--pure` is passed on the command line explicitly, and `$@` is assumed to be one argument long and pointing at a flake.

I think a good solution would be to continue to use `$1` as it is being used, but replace any flake invocation options (except maybe `--run $dump_cmd` with whatever remains in `$@` if specified. This would force users to pass `--show-trace` if they wanted that, but would only be encountered if the user specified additional arguments and thus bought into the "advanced invocation". 

I'm happy to try out a PR for this, but wanted to check in before I started work on it.

bbenne10:
Closes #135.

The fix here is simple. The arguments passed to `use_flake` are already an array.
Assuming that the first entry in that array is a flake specifier is still fine,
but this change allows us to fill the rest of the array with  other arguments to pass to `print-dev-env`.

I don't anticipate this being completely finalized, as tests should be built for this functionality (though I am unsure how right now) and the documentation should be updated to reflect the change. 

This is open for discussion as I am unsure if this is a change that the maintainers generally want.

### DIFF
--- a/direnvrc
+++ b/direnvrc
@@ -129,7 +129,7 @@ use_flake() {
     local tmp_profile_rc
     tmp_profile_rc=$("${NIX_BIN_PREFIX}nix" print-dev-env \
       --extra-experimental-features "nix-command flakes" \
-      --profile "$tmp_profile" "$flake_expr")
+      --profile "$tmp_profile" $@)
     # macos does not have realpath
     if command -v realpath >/dev/null; then
       drv=$(realpath "$tmp_profile")
--- a/direnvrc
+++ b/direnvrc
@@ -129,7 +129,7 @@ use_flake() {
     local tmp_profile_rc
     tmp_profile_rc=$("${NIX_BIN_PREFIX}nix" print-dev-env \
       --extra-experimental-features "nix-command flakes" \
-      --profile "$tmp_profile" $@)
+      --profile "$tmp_profile" "$@")
     # macos does not have realpath
     if command -v realpath >/dev/null; then
       drv=$(realpath "$tmp_profile")
--- a/README.md
+++ b/README.md
@@ -189,6 +189,24 @@ you can specify an arbitrary flake expression as parameter such as:
 use flake ~/myflakes#project
 ```
 
+### Advanced usage
+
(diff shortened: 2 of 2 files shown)


direnvrc
```diff
@@ -129,7 +129,7 @@ use_flake() {
     local tmp_profile_rc
     tmp_profile_rc=$("${NIX_BIN_PREFIX}nix" print-dev-env \
       --extra-experimental-features "nix-command flakes" \
-      --profile "$tmp_profile" "$flake_expr")
+      --profile "$tmp_profile" $@)
```
Mic92, line 132: I think this should have the desired effect too and handle quoting properly:

```suggestion
      --profile "$tmp_profile" "$@")
```

Mic92: also see shellcheck

bbenne10: TIL that POSIX handles quoting `$@` in a way that's both what I wanted all along and pretty inconsistent with how it quotes most other variables...

This should do exactly what I wanted.

bbenne10:
Updated the documentation, but I'm not sure I see a way to reuse the testing infrastructure to truly test this functionality. If anyone has any ideas, I would love to hear them

bbenne10:
Ping? Don't want to be a bother around the holidays, but am interested in resolving this issue for us :)

PR #140, opened 2022-01-29
Renew cache on changes inside a nix util directory

dvzubarev:
I often put dependencies and the package `default.nix` to the `nix` directory in the root of a project.
Now I need to edit flake.nix or flake.lock to update cache after changing files inside a nix directory.
Note that `watch_dir` was introduced in direnv `2.26.0`. So maybe it'd better to check version before invoking this.

### DIFF
--- a/direnvrc
+++ b/direnvrc
@@ -103,11 +103,24 @@ _nix_argsum_suffix() {
   fi
 }
 
+_any_file_in_dir_newer_than(){
+    local dir="$1"
+    local than="$2"
+    [[ ! -d "$dir" ]] && return 1
+    for f in "$dir/"*.nix ; do
+        [[ "$f" -nt "$than" ]] && return 0
+    done
+    return 1
+}
+
+
 use_flake() {
   flake_expr="${1:-.}"
   flake_dir="${flake_expr%#*}"
   watch_file "$flake_dir/"flake.nix
   watch_file "$flake_dir/"flake.lock
+  nix_util_dir="$flake_dir/"nix
+  [[ -d "$nix_util_dir" ]] && watch_dir "$nix_util_dir"
 
   local layout_dir profile
   layout_dir=$(direnv_layout_dir)
@@ -122,7 +135,7 @@ use_flake() {
      || "$flake_dir/"devshell.toml -nt "$profile_rc"
      || "$flake_dir/"flake.nix -nt "$profile_rc"
      || "$flake_dir/"flake.lock -nt "$profile_rc"
-     ]];
+     ]]  || _any_file_in_dir_newer_than "$nix_util_dir" "$profile_rc"
   then
     local tmp_profile="${layout_dir}/flake-profile.$$"
     [[ -d "$layout_dir" ]] || mkdir -p "$layout_dir"


direnvrc
```diff
@@ -103,11 +103,24 @@ _nix_argsum_suffix() {
...
+    return 1
+}
+
+
 use_flake() {
   flake_expr="${1:-.}"
   flake_dir="${flake_expr%#*}"
   watch_file "$flake_dir/"flake.nix
   watch_file "$flake_dir/"flake.lock
+  nix_util_dir="$flake_dir/"nix
```
Mic92, line 122: I wonder if we should just make this directory configurable? Assuming `nix` sounds like a good default to me.

bbenne10:
Is this not just the equivalent of `watch_file $(find . -name "*.nix")`?
This is what we're doing with no problems locally. I wonder if this is just not a problem for this tool to handle?

dvzubarev:
There is already [code](https://github.com/nix-community/nix-direnv/blob/f860b9dbb6708686d4ae633926090aacb4d36486/direnvrc#L118-L124) that checks whether to update cache or not.  If you use only `flake.nix` then its ok. If you use some other nix-files than nix-direnv will use "cached dev shell" even if direnv triggers reload on changed files. 
We can change the check with something like this `-n "$(find . -name "*.nix" -newer "$profile_rc" )"`. It would eliminate the need to specify nix directory, but it entails the overhead of calling find each time on reloading.

bbenne10:
Oh - I misunderstood the overall purpose here. Sorry for the noise!

Personally, I think the `find` invocation is preferable to making a decision about what files to watch based on convention.

My current setup for a particular project is a mono-repo written in 3 independent languages with `default.nix` files scattered about in subprojects. Each subproject defines a few known top-level keys (`devShellEnv`, `devShellDependencies`, and `packages`). Each attr in `packages` is a package to expose in the top-level flake. Each element in the array `devShellDependency` is a package to expose in `mkShell` and `devShellEnv` is an attr set of environment variables to overlay in the `mkShell` call, exposing them as environment variables (this doesn't handle name clashes yet. Not sure how I'd handle that if I needed it.)

This means that I'm interested in watching all of these disparate nix files wherever they are (without a `nix/` subdirectory).

No idea how common this is or how much we would want to accommodate for it, but I just wanted to provide another use-case.

dvzubarev:
> No idea how common this is or how much we would want to accommodate for it, but I just wanted to provide another use-case.

Thank you for sharing your use-case. I think we have some options to support as many as workflows as possible. 
1. Remove these [checks](https://github.com/nix-community/nix-direnv/blob/f860b9dbb6708686d4ae633926090aacb4d36486/direnvrc#L118-L124). 
2. Add check based on `find`, e.g. `-n "$(find . -maxdepth 2 -name "*.nix" -newer "$profile_rc" )"`
3. Allow an user to specify a list of watch-files. For example, before calling `use_flake` set files into some variable. Downside is that an user need to reevaluate `.envrc` each time a new file is added.
What do you think @Mic92 @bbenne10  ?

> EDITED TO ADD: Regarding version checks: My most recent PR introduces this. Maybe rebasing on that and upping the minimum version would be a good idea?

Yes if we will end up using  direnv's `watch_dir`.

bbenne10:
Either the second or third bullet point seems the best move to me. Those checks are important to make sure that the cache is only poisoned if necessary (I think. I haven't read over it deeply). 

I wonder if there's a way to query direnv for it's current watches (besides parsing `direnv status` which seems like it could be unstable between version)

dvzubarev:
> I wonder if there's a way to query direnv for it's current watches (besides parsing direnv status which seems like it could be unstable between version)

I investigated this a little as 4th option. There is env variable `DIRENV_WATCHES`, which stores watch files as base64 of zlib'ed text. It is tricky to decompress zlib in bash.

bbenne10:
There is also `direnv status  | awk '/^Loaded watch:/ { print $3 }`. I'm not saying it's good.

dvzubarev:
Ah, thanks, I didn't see it because i don't run bash hook in my `.bashrc`. I think that the `status` output differs inside `use_flake` function.

Mic92:
I think the current checks are reasonable default for small projects and I don't want to remove them for this reason to avoid boilerplate. I think instead of adding complex machinery we should have a bash function that the user can override to check whether a flake has to be reevaluated. This than hopefully cover all use cases without turning into lorri.

bbenne10:
Adding a function hook is probably the best way to handle this, imo. 
Assuming we go with it, it settles how to handle per-environment overrides.
Then the question becomes how to define this function by default.

I'd propose a two stage approach as follows (naming is subject to all sorts of changes):

```bash
_cache_is_valid() {
  if [[ ! -e "$profile"
   || ! -e "$profile_rc"
   || "$HOME/.direnvrc" -nt "$profile_rc"
   || .envrc -nt "$profile_rc"
   || "$flake_dir/"devshell.toml -nt "$profile_rc"
   || "$flake_dir/"flake.nix -nt "$profile_rc"
   || "$flake_dir/"flake.lock -nt "$profile_rc"
  ]];
  then 
    return 1;
  else 
    return 0; 
  fi
}

cache_is_valid () {
  return _cache_is_valid()
} 
```

This allows users to both override the function and to use the default implementation as part of that override. 
(I apologize for syntax mistakes above. It's 9AM on a Saturday and my 2 year old did not want me to take the 5 minutes it took to type this out)

bbenne10:
@dvzubarev: Do you want to take a stab at implementing this functionality? I'm happy to do it, if you don't want to handle the added complexity. 

There's also the possibility of both solutions being incorporated. Watching `./nix` is not a bad idea overall. It just falls short of what we need in our usage.

dvzubarev:
I was planning to implement it, but I'm kind of short of time right now. Feel free to add function hooks, I will indeed adapt to any new functionality.

bbenne10:
Please see #148 for a pretty good implementation of the general parts of this.

Issue #139, opened 2022-01-28
Cut release for 1.6.0


bbenne10:
With the last merge, we're able to use nix-direnv very well for some internal development tasks. Yay! However, I now have to explain to developers that they need at least `direnv` 2.21.3 (because `nix-direnv` assumes `shopt -s extglob` has been run, which was changed in direnv at that point release) and `nix-direnv` at `HEAD`.

It's not entirely clear from reading the documentation which version you'll get for all of the installations as of today (the nix ones get 1.5.1, but installing from source gets latest and greatest features, for instance). 

This would all be greatly simplified if:

* `nix-direnv` declared a minimum version of direnv by calling `direnv_version 2.21.3` (or later if there's some other feature I have missed that `nix-direnv` requires)
* There was a release with the latest features (namely allowing arguments to be passed to the flake invocation). I recommend 1.6.0, as this is a new feature rather than a bugfix. But 1.5.2 or 2.0.0 would be fine too. 

I'm happy to open a PR for the former, but I'm not sure I'm qualified to do the work for the latter. Please let me know what you'd prefer.

Mic92:
Can you do the pr for `direnv_version`. I will create a release afterwards.

bbenne10:
With the merge of #141, is there anything standing in the way of cutting a release?

Mic92:
done in https://github.com/nix-community/nix-direnv/releases/tag/1.6.0

bbenne10:
Thank you! This will help us quite a bit.

Mic92:
btw using 

```
if ! has nix_direnv_version || ! nix_direnv_version 1.6.0; then
  source_url "https://raw.githubusercontent.com/nix-community/nix-direnv/1.6.0/direnvrc" "sha256-FqqbUyxL8MZdXe5LkMgtNo95raZFbegFpl5k2+PrCow="
fi
```

you can also download from your fork

bbenne10:
We have some internal reasons why we didn't want to handle this that way, but we considered it. :)

PR #141, opened 2022-01-31
Declare (and enforce) a minimum direnv version

bbenne10:
2.21.3 introduces extglob to direnv's stdlib.
This means we need at least 2.21.3 to execute nix_direnv_version.
This just enforces a previously unstated dependency.

The other possible solution is to directly call `shopt -s extglob` in `nix_direnv_version`,
but I'm afraid of that having effects in other places OR of nix-direnv depending on other post-direnv-2.21.3-isms.

Part of #139

### DIFF
--- a/direnvrc
+++ b/direnvrc
@@ -5,7 +5,7 @@
 # Checks that the nix-direnv version is at least as old as <version_at_least>.
 
 if ! has direnv_version || ! direnv_version 2.21.3; then
-  exit
+  exit 1
 fi
 
 nix_direnv_version() {
--- a/direnvrc
+++ b/direnvrc
@@ -3,6 +3,11 @@
 # Usage: nix_direnv_version <version_at_least>
 #
 # Checks that the nix-direnv version is at least as old as <version_at_least>.
+
+if ! has direnv_version || ! direnv_version 2.21.3; then
+  exit
+fi
+
 nix_direnv_version() {
   declare major='1' minor='5' patch='1' # UPDATE(nix-direnv version)
 
--- a/direnvrc
+++ b/direnvrc
@@ -1,13 +1,16 @@
 # shellcheck shell=bash
 
-# Usage: nix_direnv_version <version_at_least>
-#
-# Checks that the nix-direnv version is at least as old as <version_at_least>.
+REQUIRED_DIRENV_VERSION="2.21.3"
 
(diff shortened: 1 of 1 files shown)


bbenne10:
Since #140 also introduces other versioned dependencies, I think I'm going to say that using `shopt -s extglob` inside `nix_direnv_version` is no longer viable.. The project should explicitly state what versions of direnv it supports (in the documentation as well as enforce it in the code)

direnvrc
```diff
@@ -3,6 +3,11 @@
 # Usage: nix_direnv_version <version_at_least>
 #
 # Checks that the nix-direnv version is at least as old as <version_at_least>.
+
+if ! has direnv_version || ! direnv_version 2.21.3; then
+  exit 1
```
Mic92, line 8: Cool. Could we get a nicer error message for this?

bbenne10: Sure. Do you have wording you'd prefer?
`direnv_version` already prints to stderr. Would you like me to redirect that and print our own? Or would you prefer both the default message and a follow up to clarify it?

bbenne10: Well, I took a stab while github was down. Please let me know what you think.

Mic92: Yep. that's better!

mweinelt: Where is the `has` command coming from? The direnvrc fails to load for me in zsh with this change.

```
❯ zsh
has: command not found
nix-direnv: base direnv version is older than the required v2.21.3.
❯ direnv --version
2.30.3
❯ cat ~/.zshrc
[...]
eval "$(/nix/store/gx80b9p7xdyjmbkns86zl8kkkaz5bfsl-direnv-2.30.3/bin/direnv hook zsh)"
source /nix/store/s26xz2577x2a74zd460r84m7axzjz236-nix-direnv-1.6.0/share/nix-direnv/direnvrc
[...]
```

bbenne10: `has`is documented in the direnv stdlib here: https://direnv.net/man/direnv-stdlib.1.html#codehas-ltcommandgtcode

I'm not sure why this isn't working for you. It might be worth opening an issue so that we can investigate this indepdently?

mweinelt: I was holding it wrong. I included nix-direnv inside my zshrc instead of inside the direnvrc.

bbenne10: Glad you got it sorted.

PR #145, opened 2022-02-14
DRAFT: Allow user-driven cache invalidation 

bbenne10:
Starts work on allowing user-driven cache invalidation as discussed in #140.
Still needs documentation updates.
Opening as draft to discuss approach. 
Paging @dvzubarev and @Mic92 in particular.

PR #148, opened 2022-02-21
Allow watching of arbitrary files for nix flake updates

polygon:
For more complex flake projects that branch out into other nix-files, it is currently not possible to trigger a rebuild of the environment when one of these files changes. Even if they are added to the `direnv` watches using watch_file, `use_flake` will check only some pre-determined files and otherwise will re-use the cached environment resulting in an out-of-sync environment.

This is a first suggestion on how an approach to watching multiple files might look and will likely require some re-work. Essentially, instead of using `watch_file <file>`, you now write `nix_watches+=(<file1> <file2>)`. The functionality was changed such that all the files in that array are now checked and whenever one of them changed, it will correctly trigger rebuilding the environment.


Mic92:
cc @[bbenne10](https://github.com/bbenne10)

bbenne10:
@Mic92: I suppose the question is whether we want a difference between invalidation of nix cache independent of the direnv cache. This would work fine for my use case and is a lot simpler than the PR I've opened since my approach uses functions rather than a variable. Simplicity, clarity, and aligning the nix cache and the direnv cache are all good things and so I am all for this approach over the one taken in #145.

The only trouble that I see with this is that it doesn't seem to touch `use_nix`, which (imo) it should. That should be trivial.

(As an aside: What do we think of maybe overriding `watch_file`? Doing so would allow us to unify the nix-direnv "watches" with the direnv watches (by simply dispatching to the original function) *and* would provide new users a discoverable and expected entry point that's already documented and utilized by the upstream. I don't want to cause more churn, but that seems like a pretty good solution to me?)

polygon:
@bbenne10: Thanks for the feedback. Sorry, I guess we solved a similar problem at the same time. I haven't seen your PR before opening this one. In any case, I think the function approach would still be compatible with this.

Indeed, I was going for a minimum viable version to show at first. I've added support for `use_nix` now. I also like the idea of overriding `watch_file` and implemented this. I've had to add a check if the function was already renamed because for some reason, this file is sourced twice for me.

bbenne10:
No need to be sorry for solving the sameproblem but better. I believe that your usecase is the same as mine overall and that overriding `watch_file` is a clever solution. If `use_flake` or `use_nix`is called, we unify the concepts of thr two unique caches. If not, nothing unexpected happens (presuming noone else reads `nix_watches`). Sounds like  a win to me :)

Im hoping to get feedback from @dzubarev on one or both of these PRs soon, but I understand thatlife happens and we may need to move forward wothout that.

bbenne10:
I might suggest an update to the readme describing this functionality as well.

bbenne10:
Thinking on it - if we're overriding `watch_file` we may also need to override `watch_dir` to keep everything consistent, no?

direnvrc
```diff
@@ -111,6 +111,15 @@ _nix_argsum_suffix() {
   fi
 }
 
+if [[ ! $(declare -f orig_watch_file) ]]; then
+  eval orig_"$(declare -f watch_file)"
+fi
+
+watch_file() {
```
Mic92, line 118: I wonder if it could problems down the line if we start override this function. It reminds a bit of rails monkey-patching ruby. cc @zimbatm

zimbatm: Is it possible to rename it to nix_direnv_watch_file ? I would prefer not to have to sift through watch_file implementations in the upstream bug reports.

polygon: Works for me as well. I named it nix_direnv_watch_file now.

polygon:
Function is called `nix_direnv_watch_file` now to not override any existing functionality. Open for other name suggestions. If everyone is fine with this name, it should be ready now.

bbenne10:
I think that some documentation is in order for this, but otherwise I'm happy to sign off on it.

polygon:
I've replaced the `Manually re-triggering evaluation` chapter in the readme. Please have a look.

bbenne10:
Lgtm

direnvrc
```diff
@@ -111,6 +111,11 @@ _nix_argsum_suffix() {
   fi
 }
 
+nix_direnv_watch_file() {
+  watch_file $@
```
Mic92, line 115: to fix the shellcheck error:

```suggestion
  watch_file "$@"
```

polygon: Done

Mic92:
Thanks!

PR #154, opened 2022-04-03
Convert ci tests to nix flake app test-runner

typetetris:
Turned the CI tests into an app of the flake, so they are accessible via the modern `nix` ui.

If gets rid of the `-I nixpkgs=https://github.com/NixOS/nixpkgs/archive/nixpkgs-unstable.tar.gz` in the
test invocation, which I think of as a plus.

If you don't like it, just close the pull request.


bbenne10:
Happy to merge a clean up to the testing infrastructure! Thanks for your help :)

Issue #138, opened 2022-01-13
DOCS: make it clear whether it is appropriate to track the contents of the .direnv directory with `git`


hab25:
It isn't obvious to me because there is conflicting evidence:

Evidence that makes one think it should be tracked:
- `.direnv` is created in the project directory by default.
- The tool does not touch the project's `.gitignore`. Neither the tool nor the documentation suggest modifying `.gitignore`.
- Having a shared cache could save some development time. 

Evidence that makes one think it should NOT be tracked:
- The files make references to OS/architecture (e.g. `export $'NIX_BINTOOLS_WRAPPER_TARGET_HOST_x86_64_unknown_linux_gnu'=1;`) and the nix store (e.g. `export $'out'=$'/nix/store/dj0qni0mardvd6mjr5a57ajayfpkfsa4-nix-shell'`). This makes one think that this cache could cause problems for others that differ in at least one of these aspects.
- [This documentation section](https://github.com/nix-community/nix-direnv#storing-direnv-outside-the-project-directory) suggests some problems. Aren't remote `git` repositories technically also backups?

Mic92:
To reduce friction I would actually suggest users to put `.direnv` in their global ~/.gitignore.
https://github.com/Mic92/dotfiles/blob/351ff0915b025a8c31163726591359ca22d7eb9e/.gitignore#L8
I could be still added on a project base, but not every project has .envrc upstream especially in the context of nix.

bbenne10:
I use the linked documentation section and don't track `.direnv` at all.
The whole point is that these environment are fully reproducible, so reproducing them is what we should do. 

I think the documentation section doesn't strongly enough make a case for why it is there (even if it is useful!) or make a stance either way on the tracking of `.direnv`. I'll leave this open until I can get around to updating the docs to reflect the stance outlined here.

PR #157, opened 2022-04-07
Doc updates

bbenne10:
I've gone ahead and reworked some of our README.
In particular, I want to address #138 here, but I felt I should take the time to clarify some other parts too.

They're all nicely squashed up into singular commits, so we can just cherry pick only d3651bd if the rest of the PR is overreach.

### DIFF
--- a/README.md
+++ b/README.md
@@ -258,41 +258,6 @@ nix_direnv_watch_file mod1.nix mod2.nix
 use flake
 ```
 
-## Shell integration
-
-To quickly add a `default.nix`/`flake.nix` to a project you can put the following snippets in your `.bashrc`/`.zshrc`:
-
-```bash
-nixify() {
-  if [ ! -e ./.envrc ]; then
-    echo "use nix" > .envrc
-    direnv allow
-  fi
-  if [[ ! -e shell.nix ]] && [[ ! -e default.nix ]]; then
-    cat > default.nix <<'EOF'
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -258,41 +275,6 @@ nix_direnv_watch_file mod1.nix mod2.nix
 use flake
 ```
 
-## Shell integration
-
```
Mic92, line 262: ## Shell integration

See the [wiki](https://github.com/nix-community/nix-direnv/wiki/Shell-integration) for helpers to quickly setup a direnv setup in a new project.

README.md
```diff
@@ -258,41 +275,6 @@ nix_direnv_watch_file mod1.nix mod2.nix
 use flake
 ```
```
Mic92, line 277:
```suggestion

## Shell integration

See the [wiki](https://github.com/nix-community/nix-direnv/wiki/Shell-integration) for helpers to quickly setup a direnv setup in a new project.
```

bbenne10:
I wonder if using the wiki is a good idea for other bits in the readme?
In particular, the snippet on storing direnv layout directory outside the project directory.

I think the idea is both good and very helpful, but it is not nix-direnv specific.
Maybe it is better placed in the direnv wiki?

Mic92:
`direnv_layout_dir` might be documented in direnv.

PR #160, opened 2022-04-14
Fix nix_direnv_version to properly handle major version bumps

### LINKED ISSUE #159: Newer version incorrectly reported as older
Hello, I have `.envrc` setup thusly:
```
if has nix; then
  if nix show-config | grep experimental-features | grep -q flakes; then
    if ! has nix_direnv_version || ! nix_direnv_version 1.6.0; then
      source_url "https://raw.githubusercontent.com/nix-community/nix-direnv/1.6.0/direnvrc" "sha256-FqqbUyxL8MZdXe5LkMgtNo95raZFbegFpl5k2+PrCow="
    fi
    
    use flake
  else
    use nix
  fi
fi
```
and if I enter this directory with a newer nix-direnv version, this happens:
```
direnv: using flake
direnv: nix-direnv: using cached dev shell
nix-direnv: error current version v2.0.0 is older than the desired version v1.6.0
direnv: loading https://raw.githubusercontent.com/nix-community/nix-direnv/1.6.0/direnvrc (sha256-FqqbUyxL8MZdXe5LkMgtNo95raZFbegFpl5k2+PrCow=)
direnv: using flake
```

Is this an expected behaviour?

bbenne10:
Closes #159

Just make sure that the major version is the same when checking the minor and that major and minor are the same when checking patch.

### DIFF
--- a/direnvrc
+++ b/direnvrc
@@ -44,6 +44,10 @@ nix_direnv_version() {
   [[ $1 =~ ^([^+-.]*)(\.?)([^+-.]*)(\.?)([^+-]*)(-?)([^+]*)(\+?)(.*)$ ]]
   declare -a ver; ver=("${BASH_REMATCH[@]:1}")
 
+  req_major=${ver[0]}
+  req_minor=${ver[2]:=0}
+  req_patch=${ver[4]:=0}
+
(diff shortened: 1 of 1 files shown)


Mic92:
I also noticed that.

Issue #168, opened 2022-04-25
Error when passing extra arguments to `use nix`


joncol:
Hi, after my latest NixOS system update (going to Nix 2.8, but I don't know if that's relevant), it seems I cannot pass extra arguments to `use nix` in `.envrc`.

For example, in a minimal project created using: `nix-shell -p niv --run 'niv init -b nixpkgs-unstable'`.

With `shell.nix`:
```nix
{ sources ? import ./nix/sources.nix, pkgs ? (import sources.nixpkgs { }) }:

{
  a = pkgs.mkShell { buildInputs = (with pkgs; [ hello ]); };
  b = pkgs.mkShell { buildInputs = (with pkgs; [ nyancat ]); };
}
```

and `.envrc`:
```bash
use nix -A a
```

I get the following error (when running `direnv allow` in the "project" directory):
```
direnv: loading ~/dummy/.envrc
direnv: using nix -A a
error: syntax error, unexpected invalid token, expecting ID or OR_KW or DOLLAR_CURLY or '"'

       at «string»:1:22:

            1| (import ./shell.nix).$}attribute}
             |                      ^
direnv: nix-direnv: renewed cache
direnv: export ~XDG_DATA_DIRS
```

mdarocha:
I think this is caused by the same error that was described in https://github.com/nix-community/nix-direnv/issues/162

This was fixed in https://github.com/nix-community/nix-direnv/commit/c35f4f013c2fb40a1fd221fecfce0f9c74383e4d, but unfortunately, no new version with this fix has been release yet.

Until a new nix-direnv release, you can either rollback to nix-direnv older than 2.0, or use a version from a specific commit with an overlay:
```nix
 nix-direnv = super.nix-direnv.overrideAttrs (old: rec {
    src = self.fetchFromGitHub {
      owner = "nix-community";
      repo = "nix-direnv";
      rev = "c35f4f013c2fb40a1fd221fecfce0f9c74383e4d";
      sha256 = "12opLFnNwC54T7RdBrCYWTuH20W2FVnNRN2MGIZkCpw=";
    };
  });
```

Disclaimer: This fixed my usecase with broken `use nix -p <list of packages>`. I'm not 100% sure this will fix your issue

bbenne10:
I'm not ENTIRELY sure that the fix mentioned above will resolve the issue. 
#162 was problematic in that the parsing fell back to an incorrect default case (`-*` rather than `*`).
[...]

PR #172, opened 2022-04-27
Add missing parameter to import expression

joncol:


joncol:
This seems to fix #168 for me at least.

Issue #190, opened 2022-05-31
Feature Request/Bug Report: Support for `--arg`/`--argstr`


m4dc4p:
Pardon me if this has already been requested, or exists, but it doesn't seem like `use nix` passes along `--arg` or `--argstr`. Sometimes its useful to pass those arguments to `shell.nix`.

To demonstrate, first use this `shell.nix` which just prints the argument received:

```nix
{ arg ? "xxx" }:
let
  x = import <nixpkgs> {};
in 
  builtins.trace ("Got arg: ${arg}") x.bash
```

Then use this `.envrc` which passes the `SHELL_ARG` environment variable as an argument (if present) to `shell.nix`:

```bash
if [[ -n "$SHELL_ARG" ]]; then
    echo "Passing '$SHELL_ARG' to shell"
    use nix ./shell.nix --argstr arg $SHELL_ARG;
else
    echo "Not passing arg to shell"
    use nix ./shell.nix;
fi; 
```

Executing the following bash command should produce a trace showing the SHELL_ARG arg passed (`foo`), but instead you can see it's the default value (`xxx`):

```bash
$ export SHELL_ARG=foo; direnv allow
direnv: loading ~/.../.envrc
Passing 'foo' to shell
direnv: using nix ./shell2.nix --argstr arg foo
trace: Got arg: xxx
```

For completeness, using `nix-shell` directly shows the argument is printed:

```bash
$ nix-shell ./shell.nix --argstr arg foo
trace: Got arg: foo
```

bbenne10:
Closed via #191. Please open a new issue if you find issue with the resolution.

PR #191, opened 2022-05-31
Ensure extra arguments passed to shell

m4dc4p:
Extra arguments passed to `use nix` (such as `--arg` or `--argstr`)
were getting clobbered by an errant re-definition. Removed that
definition so those arguments get passed along.

Due to #190

### DIFF
--- a/direnvrc
+++ b/direnvrc
@@ -313,7 +313,6 @@ use_nix() {
   then
     local tmp_profile="${layout_dir}/flake-profile.$$"
     local tmp_profile_rc
-    local extra_args=()
 
     if [[ "$packages" != "" ]]; then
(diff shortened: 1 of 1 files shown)


m4dc4p:
It seems like this is a pretty straightforward bug fix, but maybe there was some reason for the re-definition of `extra_arguments` that I removed?

I would have added a test for this but it wasn't obvious to me how to run the existing tests or how to add a new test. The linked issue has a repro case though, if someone wants to make it a test case.

Mic92 (approved):

Mic92:
I integrated this into https://github.com/nix-community/nix-direnv/pull/194 with a test.

Mic92:
I think we can backport this. Put it in a patch release: https://github.com/nix-community/nix-direnv/releases/tag/2.1.1

PR #212, opened 2022-07-22
Fix error with strict_env

### LINKED ISSUE #128: Reference to $NIX_BUILD_TOP in `use_flake` fails if shellHook unsets it
If my `shellHook` unsets the `NIX_BUILD_TOP` env var, there's an unguarded reference to it in the `use_flake` function that will then fail.

https://github.com/nix-community/nix-direnv/blob/69bec056fb036ff526184a995f07ee7b3f0abf0b/direnvrc#L171

This results in an error

```
/nix/store/77rjgliqwa9qp2hk1nwaic0n75l01pi2-nix-direnv-1.5.0/share/nix-direnv/direnvrc:171: NIX_BUILD_TOP: unbound variable
```

#125 rewrites a lot of this code but it still has this unguarded access to `$NIX_BUILD_TOP`.

In general, nix-direnv should assume that evaluating `$profile_rc` may make arbitrary changes to the environment and so it can't rely on any env var still existing at this point.

---

This was discovered when trying to use [numtide/devshell](https://github.com/numtide/devshell). I know there's an incompatibility right now with cached runs (see https://github.com/nix-community/nix-direnv/issues/109) but it should still work on a `direnv reload` at least

amarshall:
With direnv’s `strict_env`, previously got an error

> NIX_BUILD_TOP: unbound variable

I tried to adjust the tests to catch this, but unfortunately I could not
get them to fail as-is.

Maybe fixes #128, though that also makes me question if this is the best solution to the problem, or just a workaround.

### DIFF
--- a/direnvrc
+++ b/direnvrc
@@ -100,7 +100,7 @@ _nix_import_env() {
   # as it may be garbage collected.
   # Instead - just remove it immediately.
   # Use recursive & force as it may not be empty.
-  if [[ "$NIX_BUILD_TOP" == */nix-shell.* && -d "$NIX_BUILD_TOP" ]]; then
+  if [[ -n "${NIX_BUILD_TOP+x}" && "$NIX_BUILD_TOP" == */nix-shell.* && -d "$NIX_BUILD_TOP" ]]; then
     rm -rf "$NIX_BUILD_TOP"
   fi
 


Mic92:
Thanks!

PR #226, opened 2022-08-27
add nix_direnv_watch_dir function

Mic92:


Mic92:
Maybe we don't need this if we just watch all files by default ...

bbenne10:
I think a different name would suit this function better, but I am fine with merging it until I get an MR open that handles watching all nix files by default and comparing against `nix_watches` in a faster manner.

Maybe `nix_direnv_watch_all_nix_files_in_dir` or something?

Mic92:
I think I rather want to switch using `find`, which would compare against all nix files in the repo.

Issue #235, opened 2022-09-25
Cleaning old GC roots ?


qaristote:
If I list the profiles created by `nix-direnv` in one of my environments, here's what I get:
```
$ ls .direnv
aliases                            nix-profile-pre409534.f677051b8dc.rc  nix-profile-pre410680.a8a557e233c.rc  nix-profile-pre411003.72bdd03f0d5.rc
flake-profile.246679               nix-profile-pre410602.ae1dc133ea5     nix-profile-pre410722.e14f9fb5731     nix-profile-unknown.rc
flake-profile.246679-1-link        nix-profile-pre410602.ae1dc133ea5.rc  nix-profile-pre410722.e14f9fb5731.rc
nix-profile-pre409534.f677051b8dc  nix-profile-pre410680.a8a557e233c     nix-profile-pre411003.72bdd03f0d5
```
These profiles are all pointed to by a GC root so will not get cleaned after a `nix-collect-garbage -d`, yet I have strong reasons to believe most of them are actually profiles for older versions of my environment (they don't appear again if I delete them, their `buildInputs` do not match the current environment, I didn't find any part of `nix-direnv`'s code that deletes old roots) and are thus mostly useless.

Is there a way to have `nix-direnv` automatically delete these old profiles ? I could of course do something like `rm .direnv/nix-profile*` followed by `direnv reload` but maybe there's a better way ?

dschrempf:
I was surprised that you have old configurations in your `.direnv` directories. For me this is not the case, it only stores the newest. (However, I use `use flake`).

Snektron:
I have this same issue with `use flake`

bbenne10:
I'm still working on some performance improvements for #227 in my downtime (and life is busier than ever right now). 
I'm happy to discuss and probably merge a pull request that deals with this shortcoming.

bbenne10:
@qaristote's MR is now merged and I'm going to call this done! Thanks again, @qaristote :)

PR #246, opened 2022-10-19
remove old gcroots when the environment is updated

qaristote:
As discussed in #235, nix-direnv currently does not remove the profiles and GC roots created for older version of the environment. This PR aims to fix this.

It adds a function `_nix_clean_old_gcroots` that is called before updating the environment and removes all the `nix-{flake,profile}*` from the local `.direnv` directory as well as the corresponding GC roots. Since this function reuses the code of `_nix_add_gcroots` for escaping path, this code is now run by calling the `_nix_strip_escape_path` function.

I only tested this PR with the current test-suite as I didn't take the time to understand what the tests were doing. Let me know if you want me to add some tests.

### DIFF
--- a/tests/test_gc.py
+++ b/tests/test_gc.py
@@ -64,7 +64,9 @@ def test_use_nix(direnv_project: DirenvProject) -> None:
     direnv_project.setup_envrc("use nix")
     common_test(direnv_project)
 
-    direnv_project.setup_envrc("use nix --argstr 'echo Executing hijacked shellHook.'")
+    direnv_project.setup_envrc(
+        "use nix --argstr shellHook 'echo Executing hijacked shellHook.'"
+    )
     common_test_clean(direnv_project)
 
 
--- a/tests/test_gc.py
+++ b/tests/test_gc.py
@@ -64,8 +64,7 @@ def test_use_nix(direnv_project: DirenvProject) -> None:
     direnv_project.setup_envrc("use nix")
     common_test(direnv_project)
 
-    # --pure here is just a way to make sure the environment changes
-    direnv_project.setup_envrc("use nix --pure")
+    direnv_project.setup_envrc("use nix --argstr 'echo Executing hijacked shellHook.'")
     common_test_clean(direnv_project)
 
 
@@ -81,7 +80,6 @@ def test_use_flake(direnv_project: DirenvProject) -> None:
     for symlink in inputs:
         assert symlink.is_dir()
 
-    # --impure here is just a way to make sure the environment changes
     direnv_project.setup_envrc("use flake --impure")
     common_test_clean(direnv_project)
 
--- a/tests/test_gc.py
+++ b/tests/test_gc.py
@@ -56,8 +56,6 @@ def common_test_clean(direnv_project: DirenvProject) -> None:
     profiles = [f for f in files if not f.match("*.rc")]
     if len(rcs) != 1 or len(profiles) != 1:
         print(list(files))
-    print(list(files))
-    assert False
     assert len(rcs) == 1
     assert len(profiles) == 1
 
@@ -83,8 +81,6 @@ def test_use_flake(direnv_project: DirenvProject) -> None:
     for symlink in inputs:
         assert symlink.is_dir()
 
-    files = list((direnv_project.dir / ".direnv").iterdir())
-    print(files)
     # --impure here is just a way to make sure the environment changes
     direnv_project.setup_envrc("use flake --impure")
     common_test_clean(direnv_project)
--- a/tests/test_gc.py
+++ b/tests/test_gc.py
@@ -55,7 +55,7 @@ def common_test_clean(direnv_project: DirenvProject) -> None:
     rcs = [f for f in files if f.match("*.rc")]
     profiles = [f for f in files if not f.match("*.rc")]
     if len(rcs) != 1 or len(profiles) != 1:
-        print(list(files))
+        print(files)
     assert len(rcs) == 1
     assert len(profiles) == 1
 
--- a/tests/test_gc.py
+++ b/tests/test_gc.py
@@ -38,9 +38,29 @@ def common_test(direnv_project: DirenvProject) -> None:
     assert "Executing shellHook." in out2.stderr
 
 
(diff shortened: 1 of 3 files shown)


bbenne10:
Thank you! This looks pretty good.

I think I have two comments:

* I would like to see some tests for this, but I don't QUITE know how I'd go about implementing them as you'd need to change the environment mid execution. Maybe a simple test that tests the removal functionality itself by setting up an appropriate directory structure and then sourcing `direnvrc` and executing the removal function? Not sure that's enough though.
* I had considered adding a gcroot to first encompass removing the *other* gcroots. I wonder if making it a semi-atomic procedure (by moving the gcroot removal logic into `_nix_direnv_add_gcroot`) has any advantages? Or is that too difficult because of the way the loop on (new) line 224 is called?

qaristote:
> I had considered adding a gcroot to first encompass removing the _other_ gcroots. I wonder if making it a semi-atomic procedure (by moving the gcroot removal logic into `_nix_direnv_add_gcroot`) has any advantages? Or is that too difficult because of the way the loop on (new) line 224 is called?

The reasons I created a new function instead of changing `nix_add_gcroot` are 1) `_nix_clean_old_gcroots` also deletes all `nix-{flake,profile}*.rc` files but the new `.rc` file is added before `_nix_add_gcroot` is called and 2) this is probably the simplest way to go.
But yeah making the GC roots management semi-atomic as you described would have the advantage that if something fails while updating the environment the old environment does not get deleted (although even if it were it would be easy to rebuild as long as the GC hasn't been run). And while this would require more work it would certainly be possible.

I don't know whether I'll put time into this so one option would be to just merge this once more tests are added and open an issue discussing the need for this to be semi-atomic.

qaristote:
[24d0af4](https://github.com/nix-community/nix-direnv/pull/246/commits/24d0af43fafe48b90038875ed6dbf683587ec65c) adds a test that checks that `.direnv/` indeed gets cleaned. It does so by resetting `.envrc` to use a pure shell, effectively updating the environment, and then checking that `.direnv` contains the expected number of files.
Note that I don't use flakes so I assumed the number of files I was seeing during the tests was the expected one.

Here are other tests that could / should be implemented:
- [x] test that `.direnv` gets cleaned; 
- [ ] test that `.direnv/flake-inputs` gets cleaned;
- [ ] test that `/nix/var/nix/gcroots/per-user/$USER` gets cleaned (harder, as it requires getting the user and computing the escaped path).

qaristote:
> It does so by resetting `.envrc` to use a pure shell

Actually the `--pure` trick seems to not be working for `use_flake` : somehow if I use `use-flake -ignore-environment` only a new `.rc` file gets added, but no new flake profile is. I'll have to look into that.

qaristote:
I'm now forcing a change to the environment in the `use_nix` case by passing the `shellHook` as an argument, which feels like less of a hack.

 I don't know how to do the same for the `use_flake` case though as I don't use flakes. It seems `nix print-dev-env` has options to manually override the inputs which would also help to deal with the cleaning of `flake-inputs`. I'm thinking of adding a local input to the flake which would specifiy the `shellHook` for example and would come in two versions, but again I don't know how to do that so suggestions are welcome.

tests/test_gc.py
```diff
@@ -38,10 +38,35 @@ def common_test(direnv_project: DirenvProject) -> None:
...
+        print(list(files))
+    assert len(rcs) == 1
+    assert len(profiles) == 1
+
+
 def test_use_nix(direnv_project: DirenvProject) -> None:
     direnv_project.setup_envrc("use nix")
     common_test(direnv_project)
 
+    direnv_project.setup_envrc("use nix --argstr 'echo Executing hijacked shellHook.'")
```
bbenne10, line 67: Shouldn't `argstr` swallow *two* arguments here? The name of the argument and the value to pass in?
I'd expec this to be:

```
direnv_project.setup_envrc("use nix --argstr shellHook 'echo Executing hijacked shellHook'")
```

I haven't tested, but I'm suspicious of why tests are passing with this (seemingly) obvious flaw.
Am I overlooking something?

bbenne10: Or does the exact value of the `argstr` argument simply not matter as it is used only as a "cache buster" here?

I'd still expect `nix print-dev-env` to fail with this invocation though,as I think this will end up creating a value of `extra_args`in `use_nix` that looks like:

```bash
extra_args=("--argstr "'echo Executing hijacked shellHook.'" "")
```

Which I would expect to fail.

qaristote: Yeah that was a mistake. I assume the test didn't fail because the tests that come afterwards only check that a specific number of files are present in `.direnv/` and would be verified if the environment weren't reloaded at all (because reloading the environment itself failed).

bbenne10: This indicates to me that the test itself is flawed since we don't notice failure cases. I'm not sure what to do about this (testing for things like nix-direnv is always a bit difficult due to the stateful nature of everything that is involved). Making things worse, I can't run the tests locally (unsure why but this is a longstanding issue that I have yet to figure out :/) to work on the flakes. I think the `--impure` flag should do, but I'm not sure how much I trust the test's current passing state given that there were no failures in the intermediate state

qaristote: I would argue that the test is not flawed since it wasn't noticing a failure of the test itself. The point of the test is to test that the code is correct, not that the test itself is correct. But an easy addition would be to also check the exit code of the various commands (I assume the exit code would have been non-zero with `use_nix --argstr "echo Executing hijacked shellHook."`).

Overall, assuming the tests I wrote are correct, we can be confident that `.direnv/` gets cleaned. But we can't say anything about `.direnv/flake-inputs` and `/nix/var/nix/gcroots`, so I wouldn't trust the new code too much as well.

qaristote:
I'm not sure I'll be contributing much more to this for now, so [here](https://github.com/qaristote/nix-latitude7490/blob/922a9f3820f94f84cee9ebe7792e0ce4e1d08bac/home/config/direnv.nix)'s my temporary workaround if anyone is interested: I just create a systemd service that deletes the old gcroots daily.

bbenne10:
This looks good. The few things I see that I want fixed in this are either nitpicks or very much not your doing. Thank you for your contribution!

PR #254, opened 2022-11-09
fix(direnvrc): log when nix-shell not found

### LINKED ISSUE #253: Preflight logs "nix version is older than the required" when nix is not installed
# How to reproduce
Use nix-direnv on an environment without nix (can't run `nix-shell`).

# Current Behavior
nix-direnv outputs:
```
nix-direnv: nix version is older than the required 2.4.
```

# Expected Behavior
nix-direnv outputs:
```
nix-direnv: command not found: nix-shell.
```

aabccd021:
Closes #253


Mic92:
Thanks!

PR #262, opened 2022-11-27
update nixos example for 22.11

Mic92:


Mic92:
Yes it was deprecated but still works with a warning.

PR #269, opened 2022-12-05
ensure XDG_DATA_DIRS contains unique entries

mkhl:
after importing from the nix environment
we currently concatenate the new and old values of XDG_DATA_DIRS,
which means if the nix environment reused or recreated the old values
they are now present twice.
XDG_DATA_DIRS is specified to be a "set",
and some software (GNOME for example) assume that its entries are unique.

this change reconstructs XDG_DATA_DIRS
by looping over the new and old entries
and adding the new ones among them.
it normalizes the entries by removing trailing slashes
to make duplicate detection a bit easier.

it could instead loop over only the old entries and append them to the new ones,
that would save one line of code but make duplicate detection a bit more involved.

looping only over the new entries would make invert their order.

### DIFF
--- a/direnvrc
+++ b/direnvrc
@@ -113,7 +113,16 @@ _nix_import_env() {
   _nix_export_or_unset TMPDIR "$old_tmpdir"
   _nix_export_or_unset TEMP "$old_temp"
   _nix_export_or_unset TEMPDIR "$old_tempdir"
-  export XDG_DATA_DIRS=$XDG_DATA_DIRS${old_xdg_data_dirs:+":"}$old_xdg_data_dirs
+  local new_xdg_data_dirs=${XDG_DATA_DIRS:-}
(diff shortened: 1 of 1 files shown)


bbenne10:
I think the scope on this feels off.
We should ensure that this fix applies to anywhere we handle concatenation (indicating it should be a function to ease reuse), but we should also ask "Why is Gnome handling this poorly?" And maybe ask them to fix their assumption that `XDG_DATA_DIRS` is a set.

I don't mind it being merged (@Mic92 took care of that :P), but it feels like there's a bigger issue here.

Mic92:
Since we are taking away the opportunity from the setup hooks merge XDG_DATA_DIRS correctly, we need to have the semantics in nix-direnv instead. If we have other variables that need this treatment, we can have a function later on.

PR #271, opened 2022-12-06
Don't track $nixfile if it is empty

### LINKED ISSUE #270: use nix -p foo bar creates erroneous file watch
If I use e.g `use nix -p poetry vault` without `shell.nix` or `default.nix` in specific directory, then nix-direnv creates erroneous watch against $PWD, making every change to $PWD trigger direnv reload.

I believe this is the culprit: https://github.com/nix-community/nix-direnv/blob/22b5729f266a6d59746fdb12eb3303bb2de63896/direnvrc#L341 (`"$nixfile"` variable ending up empty)

If you do `watch_file ""` (without `use nix` etc.), then exactly the same behaviour will appear.

direnv status output:

```
direnv exec path /Users/mark/.nix-profile/bin/direnv
DIRENV_CONFIG /Users/mark/.config/direnv
bash_path /nix/store/49wn01k9yikhjlxc1ym5b6civ29zz3gv-bash-5.1-p16/bin/bash
disable_stdin false
warn_timeout 5s
whitelist.prefix []
whitelist.exact map[]
Loaded RC path /home/mark/test/.envrc
Loaded watch: ".envrc" - 2022-12-06T18:33:52+01:00
Loaded watch: "../.local/share/direnv/allow/3834b5cd6a767425b583e6c2481bfaa6064fd7a50148c2e1fa8329de5ae4bb6d" - 2022-12-06T18:36:40+01:00
Loaded watch: ".envrc.local" - 2022-12-06T18:49:47+01:00
Loaded watch: "../.direnvrc" - 1970-01-01T01:00:00+01:00
Loaded watch: "../.config/direnv/direnvrc" - 2022-12-06T18:38:10+01:00
Loaded watch: "shell.nix" - 1970-01-01T01:00:00+01:00
Loaded watch: "default.nix" - 1970-01-01T01:00:00+01:00
Loaded watch: "." - 1970-01-01T01:00:00+01:00
Loaded RC allowed false
Loaded RC allowPath
Found RC path /home/mark/test/.envrc
Found watch: ".envrc" - 2022-12-06T18:33:52+01:00
Found watch: "../.local/share/direnv/allow/3834b5cd6a767425b583e6c2481bfaa6064fd7a50148c2e1fa8329de5ae4bb6d" - 2022-12-06T18:36:40+01:00
Found RC allowed true
Found RC allowPath /Users/mark/.local/share/direnv/allow/3834b5cd6a767425b583e6c2481bfaa6064fd7a50148c2e1fa8329de5ae4bb6d
direnv: loading ~/home/mark/test/.envrc
direnv: loading ~/home/mark/test/.envrc.local
direnv: using nix
direnv: nix-direnv: using cached dev shell
```

bbenne10:
Should resolve #270.

I want to make sure the tests pass before we merge, but the fix is pretty trivial

### DIFF
--- a/direnvrc
+++ b/direnvrc
@@ -337,7 +337,7 @@ use_nix() {
   done
 
   nix_direnv_watch_file "$HOME/.direnvrc" "$HOME/.config/direnv/direnvrc" ".envrc"
-  if [ -e "$nixfile" ]; then
+  if [ -f "$nixfile" ]; then
     nix_direnv_watch_file "$nixfile"
   fi
 
--- a/direnvrc
+++ b/direnvrc
@@ -336,9 +336,10 @@ use_nix() {
     esac
   done
 
-  # nixfile may be empty,
-  # but nix_direnv_watch_file checks for existence before adding to watches
-  nix_direnv_watch_file "$HOME/.direnvrc" "$HOME/.config/direnv/direnvrc" .envrc "$nixfile"
+  nix_direnv_watch_file "$HOME/.direnvrc" "$HOME/.config/direnv/direnvrc" ".envrc"
+  if [ -e "$nixfile" ]; then
+    nix_direnv_watch_file "$nixfile"
+  fi
 
   local need_update=0
   local file=
--- a/tests/direnv_project.py
+++ b/tests/direnv_project.py
@@ -1,10 +1,11 @@
 #!/usr/bin/env python3
 
-from dataclasses import dataclass
 import shutil
+import textwrap
+from dataclasses import dataclass
+from pathlib import Path
 from tempfile import TemporaryDirectory
 from typing import Iterator
-from pathlib import Path
 
 import pytest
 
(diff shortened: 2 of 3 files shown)


bbenne10:
I obviously don't completely understand the test harness as well as I had hoped. I'll get on fixing those up this evening.

direnvrc
```diff
@@ -336,9 +336,10 @@ use_nix() {
     esac
   done
 
-  # nixfile may be empty,
-  # but nix_direnv_watch_file checks for existence before adding to watches
-  nix_direnv_watch_file "$HOME/.direnvrc" "$HOME/.config/direnv/direnvrc" .envrc "$nixfile"
+  nix_direnv_watch_file "$HOME/.direnvrc" "$HOME/.config/direnv/direnvrc" ".envrc"
+  if [ -e "$nixfile" ]; then
```
Mic92, line 340: Can this be something that is not a file?

```suggestion
  if [ -f "$nixfile" ]; then
```

mikroskeem: Maybe symlink?

bbenne10:
During testing, I realized that we're also tracking other files that may not exist (like `default.nix` or `shell.nix` in this particular case).

This was the reason I added the file existence check to `nix_direnv_watch_file` in the first place, but I'm not sure how to add that check back in without regressing in performance :/

PR #275, opened 2022-12-17
Only archive flake inputs excluding `self` when registering flake inputs as gcroots

Ma27:
NOTE: this should probably include a check on whether Nix is new enough to support lazy trees, but for now that's just WIP.

From my understanding we only need the flake's input path itself because Nix would copy it into the store anyways when evaluating a flake. However that's not the case anymore when using the lazy-trees branch (so neither Nix master nor a Nix release).

This change came from the motivation to not have my flakes in my store unless I explicitly need it. This is because flakes contain secrets in my case (these are encrypted via git-crypt, so not leaked in the remote repo), but they're readable by my user in my git checkout. When copying this flake into my local store, my local store makes these secrets available for each user & process on my system which is not nice (even though it's not the end of the world in my case because it's a single-user machine). However I'd like to avoid that and now I can by using lazy-trees, this change is just the last bit to achieve this.

### DIFF
--- a/direnvrc
+++ b/direnvrc
@@ -230,6 +230,7 @@ use_flake() {
     # also add garbage collection root for source
     local flake_input_paths
     mkdir "$flake_inputs"
+    # shellcheck disable=SC2016
     flake_input_paths=$("${NIX_BIN_PREFIX}nix" eval \
       --impure --json --expr \
       'builtins.mapAttrs (_: { outPath, ... }: "${outPath}") (builtins.getFlake (toString ./.)).inputs' \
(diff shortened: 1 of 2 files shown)


Mic92:
makes sense.

bbenne10:
This adds a dependency on `jq` without adding a full-fledged dependency on it.
There's already a patch for `grep` turning into a full path for nixpkgs' `gnuGrep`, which we can use as a template for a fix, I think. (We have a similar issue with bash, but can't fix that by patching as we use whatever version we're invoked with).

Mic92:
Yeah, if possible I would prefer no dependency on jq. nix-direnv is supposed to work on bare-bone Linux/MacOS machines.

Mic92:
Seems work seems to be stalled. Just re-open when you plan to continue working on it.

Ma27:
Yeah, no time for that right now and not even using lazy trees rn anymore... if someone else wants to continue, feel free.

PR #282, opened 2023-01-12
Prefer `packages` over `nativeBuildInputs`

hannseman:
Since https://github.com/NixOS/nixpkgs/commit/c6b62f2381571cc83c6c32ef4984fabcbb4ca892 we now have the `packages` attribute which at least in the nixpkgs repository seems to be preferred over `nativeBuildInputs` in the `mkShell` context. 

I also removed the empty `buildInputs` attribute. These two changes makes the template a bit more beginner-friendly in my opinion and aligns it with the documentation on `pkgs.mkShell`: https://nixos.org/manual/nixpkgs/stable/#sec-pkgs-mkShell

### DIFF
--- a/README.md
+++ b/README.md
@@ -180,7 +180,7 @@ Either add `shell.nix` or a `default.nix` to the project directory:
 { pkgs ? import <nixpkgs> {}}:
 
 pkgs.mkShell {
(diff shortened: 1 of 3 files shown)


Mic92:
bors merge

PR #287, opened 2023-01-16
improve handling if no NIX_PATH is set

Mic92:


Mic92:
bors merge

bbenne10:
@Mic92: Just a note that these bors commands are timing out and not properly merging as I think you expect. I'm not entirely sure how to resolve (or even how to use bors :P) but thought you should at least get explicitly pinged regarding failure.

Mic92:
bors merge

Issue #305, opened 2023-02-22
Pass on nix error messages


pinselimo:
I just ran into ``nix-direnv`` failing with: "direnv: nix-direnv: nix version is older than the required 2.4."
I am running nix 2.13.2 and the issue was caused by the nix store running out of space.
This was reported by ``nix-shell`` but ignored by ``nix-direnv`` in its error reporting.

Wouldn't it be better to report the error message given by ``nix-shell`` in the check and amend it with something along the lines of "Maybe your nix version is older than the required 2.4" ?

bbenne10:
Yes - this would be a better behavior. As of now, we're invoking `nix` with options we know only exist on new enough versions and it fails because this is the first invocation of nix in the script and we just so happen to also redirect stderr here. 

I think we should instead use `nix --version` and parse the return string to get the current version, letting stderr bubble there. I'm not sure if it will fail there, but it should soon afterwards.

I probably won't have time to handle this soon, but a PR is welcome.

PR #308, opened 2023-02-23
Check nix version directly

pinselimo:
This PR resolves issue #305 by querying the nix version through ``nix --version`` and testing that.

It also adds error reporting of the current nix version in case the test fails.


bbenne10:
This LGTM, but I'd like to have @Mic92 make sure this doesn't break some other obscure usecase that's not quite tested or stated well enough. 

I personally think this is better: We're explicitly testing for the thing we're stating fails and not redirecting any other failure case out of the user's site. With flakes shipped in stable versions of nix for a while now, I (personally) don't expect too many differences between what they look like now vs when they are deemed "stable" (though I do wish that would happen sooner rather than later...)

bbenne10:
Oops. Looks like you did not run shellcheck over the file. Please do so and fix up the reported errors. (You can see the ones it reported in the CI failure logs)

pinselimo:
Thanks @bbenne10 ! Fixed it up and ran the tests locally. Had to introduce an intermediate variable.

bbenne10:
With no response from @Mic92, I'm going to merge this. We'll handle any edge cases in a new issue or PR.

pinselimo:
Thx @bbenne10! I've tried to test the regex against all abominations of a nix version string I could come up with and it held up. Let's hope it stays this way.

PR #311, opened 2023-03-01
Use nix-store --add-root instead of hard-coding gcroot path (fix for Nix 2.14+)

amarshall:
Do so by no longer hard-coding gcroot paths, and defer to Nix to manage things.

https://github.com/NixOS/nix/pull/5226

### DIFF
--- a/direnvrc
+++ b/direnvrc
@@ -147,13 +147,9 @@ _nix_add_gcroot() {
 
 _nix_clean_old_gcroots() {
   local layout_dir=$1
-  local escaped_layout_dir
-  escaped_layout_dir=$(_nix_strip_escape_path "$layout_dir")
 
   rm -rf "$layout_dir/flake-inputs/"
   rm -f "$layout_dir"/{nix,flake}-profile*
-  rm -f "/nix/var/nix/gcroots/per-user/$USER/$escaped_layout_dir"-flake--inputs*
-  rm -f "/nix/var/nix/gcroots/per-user/$USER/$escaped_layout_dir"-{nix,flake}--profile*
 }
 
 _nix_argsum_suffix() {
--- a/direnvrc
+++ b/direnvrc
@@ -127,22 +127,12 @@ _nix_import_env() {
   done
 }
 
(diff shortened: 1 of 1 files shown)


bbenne10:
I'm quite confused as to why this is failing.
I see the changes in 2.14 and understand that we need to update our handling of gcroots and profiles to account for the move.
I too am unsatisfied with the tests as they stand, but I am unsure how to test these changes in a more fine grained manner. I believe you can comment out the `nix-collect-garbage` invocations and still have valid results.

However, the main question I have right now (and I am surprised that it's taken me this long to realize this) is: Why aren't we using `nix-store --add-root`? We're simply reinventing the wheel here, I think? That'd allow us to not worry about this at all, wouldn't it? We would only need to worry about getting `$profile_rc` and `$profile` into `$layout_dir` then, I think.

bbenne10:
Is it really orthogonal? it creates the roots without us having to worry about exactly where the gcroot got created.

Regarding removal: couldnt we just remove the created files inside $LAYOUT_DIR, which would cause the symlink to dangle and then qualify that root for collection?

I tried to figure this out but couldnt get nix 2.14 no matter how I tried. The latest i got was 2.13.3 with latest nixpkgs-unstable and the nixUnstable package...

Mic92:
I have pinned the ci to nix 2.13 for now. You may want to unpin in this PR for testing: https://github.com/nix-community/nix-direnv/pull/312

bbenne10:
Had a play around with this yesterday (using the podman image released as part of the [nix-unstable-installer releases](https://github.com/nix-community/nix-unstable-installer)) and the problem is either in gcroot creation or gcroot persistence after garbage collection. The first pass seems to work well enough (after first GC, but pre second GC in the tests) but post-GC, we consistently enter the "cache invalid" portion of the branch logic. I modified direnv to print out WHY we were entering that part of the branch and it is because `$profile` is missing. I'm not sure yet why that is so, but I suspect we're not doing enough to register a gcroot.

Mic92:
@bbenne10 please do so.

PR #321, opened 2023-03-15
Suggest XDG_CONFIG_DIR rc file for direnv

smcgivern:
direnv itself suggests this path instead of ~/.direnvrc: https://direnv.net/man/direnv.1.html#files


Mic92:
bors merge

Mic92:
Thanks!

PR #329, opened 2023-04-13
Fix use nix with no nixpkgs when strict_env

amarshall:
strict_env basically does `set -e`, and this fails in that case. We handle it failing in the `if` on the subsequent line.

I’m curious if `setup_envrc` in the tests should be changed to set `strict_env` to catch this sort of thing—or maybe parameterize all the tests over both cases?

### DIFF
--- a/direnvrc
+++ b/direnvrc
@@ -256,7 +256,7 @@ use_nix() {
 
   local layout_dir path version
   layout_dir=$(direnv_layout_dir)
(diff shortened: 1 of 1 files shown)


bbenne10:
I agree that the underlying issue is an issue, but I am not sure that this does what we want here. 

Since we're capturing output rather than return code, `|| true` will set the value of

bbenne10:
I apologize for the churn. I must've accidentally clicked something without intending to?

Anyway - I think this looks fine.

We probably should set strict env in the tests, since I think it can only find *new* errors rather than *different* errors, yeah?

PR #332, opened 2023-04-14
Run tests with and without direnv’s strict_env

amarshall:
Based on discussion in https://github.com/nix-community/nix-direnv/pull/329.


bbenne10:
I think that this needs a rebase to fix up the tests and then I'm happy to merge it.

bbenne10:
And we're merged :) Thanks

PR #350, opened 2023-06-03
Apply all arguments to `nix flake archive` command

sepiabrown:
In Repl.it, I want to disable downloading the source from online so I use the following code inside `.envrc`.

`use flake .# "--override-input" "nixpkgs" "git+file:./nixpkgs?shallow=1"`

This code allows the `nix flake archive` part of `use_flake` function to respect the arguments given.

### DIFF
--- a/direnvrc
+++ b/direnvrc
@@ -222,7 +222,7 @@ use_flake() {
       --json \
       --extra-experimental-features "nix-command flakes" \
       --no-write-lock-file \
-      "$flake_dir" | grep -E -o '/nix/store/[^"]+')
+      "$@" | grep -E -o '/nix/store/[^"]+')
     for path in $flake_input_paths; do
       _nix_add_gcroot "$path" "${flake_inputs}/${path##*/}"
     done


bbenne10:
Are we sure there are no arguments that can be applied to `print-dev-env` that cannot be applied to `archive`? 

The problem I foresee with this approach is that reusing the arguments in this way could lead to some odd issues if `archive` and `print-dev-env` share only a subset of arguments. 

I have not looked into this yet, and will likely not have much time for about a month to actually do the diving.

sepiabrown:
Are there any use cases where you pass `--json`, `--profile`, or `--redirect` while using `envrc`? If not, then I see no problem!

bbenne10:
I think I believe that `--json`, `--profile`, and `--redirect` are probably useless as arguments to both `archive` and `print-dev-env` here. But I think we may also want to blacklist them rather than letting the underlying invocation fail and print obtuse error messages to the caller. Can you modify to add a switch/case which blacklists these with a useful error message? Something along the lines of (untested):

```bash
  while [[ "$#" -gt 0 ]]; do
    i="$1"
    shift

    case $i in
      --redirect)
      --json)
      --profile)
       log_status "nix-direnv: Unsupported flag found in use_flake: $i"
    esac
  done
```

This should go right in between 169 and 170 (just after `use_flake` invokes `_nix_direnv_preflight`). 
If the tests pass with this (and they should), I'll merge.

bbenne10:
@sepiabrown: Did you have intentions to come back to this?

sepiabrown:
> @sepiabrown: Did you have intentions to come back to this?

Yes! I was busy and I think I won't be able to work on it until this month.

Mic92:
@sepiabrown feel free to re-open this PR, when you got time to look into it.

PR #372, opened 2023-08-06
Warn when `use flake` receives a flag for the first argument

cydparser:
This bit me recently. Usually `use flake -L` and `use flake -L some-flake` work without errors. When an error does occur, the resulting error messages can be obscure.

### DIFF
--- a/direnvrc
+++ b/direnvrc
@@ -195,7 +195,7 @@ use_flake() {
   flake_expr="${1:-.}"
   flake_dir="${flake_expr%#*}"
 
-  if [[ "$flake_expr" =~ ^-[^#]*$ ]]; then
+  if [[ "$flake_expr" = -* ]]; then
     if [[ -n "$2" ]]; then
       log_status "nix-direnv: the first argument must be a flake expression"
     else
--- a/direnvrc
+++ b/direnvrc
@@ -195,6 +195,14 @@ use_flake() {
   flake_expr="${1:-.}"
   flake_dir="${flake_expr%#*}"
 
+  if [[ "$flake_expr" =~ ^-[^#]*$ ]]; then
+    if [[ -n "$2" ]]; then
+      log_status "nix-direnv: the first argument must be a flake expression"
+    else
+      log_status "nix-direnv: the first argument must be a flake expression. did you mean 'use flake . $1'?"
+    fi
+  fi
+
   local files_to_watch
   files_to_watch=(".envrc" "$HOME/.direnvrc" "$HOME/.config/direnv/direnvrc")
 


direnvrc
```diff
@@ -195,6 +195,14 @@ use_flake() {
   flake_expr="${1:-.}"
   flake_dir="${flake_expr%#*}"
 
+  if [[ "$flake_expr" =~ ^-[^#]*$ ]]; then
+    if [[ -n "$2" ]]; then
+      log_status "nix-direnv: the first argument must be a flake expression"
+    else
+      log_status "nix-direnv: the first argument must be a flake expression. did you mean 'use flake . $1'?"
```
bbenne10, line 202: Couldn't this be simplified to something like the below?

```bash
if [[ "$flake_expr" = "-*" ]] then
  log_status "nix-direnv: First argument appears to be an option. Did you mean 'use flake . $@'?"
fi
```

Both of our approaches preclude `$flake_expr` being a file starting with `-` (which is a case that I honestly don't care *that* much to support explicitly). I'd prefer to avoid regexes if possible for simplicity's sake and I'm not sure that there is much value in special casing the 2 argument case?

cydparser: > if [[ "$flake_expr" = -* ]]; then

Yes, that is probably good enough.

> log_status "nix-direnv: First argument appears to be an option. Did you mean 'use flake . $@'?"

I don't think we can easily provide a suggestion when there is more than one argument. E.g. without the special casing, `use flake -L --fallback github:foo/bar` would result in a very wrong suggestion: `Did you mean 'use flake . -L --fallback github:foo/bar'?`

bbenne10: Sorry it has taken me a bit to reply, but it isn't clear to me how these inputs differ meaningfully.

We already will fail if the flake to build is embedded in the invocation string deeper than `$1`.
I understand that the suggestion above will fail also, but I'm not to concerned about being wrong (it is just a suggestion, not something we've guaranteed to work and we have provided the reason why already so users can interpret the failure warning independent of the suggestion).

Is `-L --fallback` somehow dangerous in this case? I don't see it, but I make light use of the varied nix flags and want to make sure before I form a strong opinion.

cydparser: > I understand that the suggestion above will fail also, but I'm not to concerned about being wrong

I personally find wrong suggestions frustrating.

> Is -L --fallback somehow dangerous in this case?

No, they are not dangerous. If one tried the wrong suggestion in my previous comment, they would see something like the following:
```
error: unexpected argument 'github:foo/bar'
Try '/nix/store/ywlpcc5zikgjhgcv09avzyz7p1sd0f3b-nix-2.15.1/bin/nix --help' for more information.
./.envrc:1389: pop_var_context: head of shell_variables not a function context
environment:1: pop_var_context: head of shell_variables not a function context
direnv: error exit status 1
```

The error message now suggests `nix --help`, which isn't very useful to someone who is trying to use `nix-env`.

PR #388, opened 2023-09-23
update README... again

Gerg-L:
The nixpkgs direnv module automatically sources nix-direnv so no need to add it to the users direnvrc
and the persistDerivations option was removed as it's no longer needed
:)


Mic92:
Thanks!

Issue #393, opened 2023-10-09
_nix_direnv_force_reload: unbound variable


lilyball:
With the current latest nix-direnv (a929b6bede9e6a22d8182847c4045f69728f24bd) and `strict_env` set, nix-direnv fails with

```
/nix/store/1wl5gscb1rwki9a44v65292zcbxnsjjg-nix-direnv/share/nix-direnv/direnvrc:237: _nix_direnv_force_reload: unbound variable
```

The code in question is testing `-z "$_nix_direnv_force_reload"`, but the variable is not set and `strict_env` turns on the option that errors out for unset variables. So either this variable needs to be unconditionally set, or any code that references it needs to use the form that handles unset variables (i.e. `${_nix_direnv_force_reload-}`).

PR #394, opened 2023-10-09
fix _nix_direnv_force_reload not beeing bound

Mic92:
fixes https://github.com/nix-community/nix-direnv/issues/393


Mic92:
@mergify queue

PR #401, opened 2023-10-31
Fix Manual Reload script with non-standard layout dir

### LINKED ISSUE #400: Manual reloading assumes default direnv_layout_dir
The `nix-direnv-reload` script that's created when `nix_direnv_manual_reload` is enabled assumes that the `direnv_layout_dir` function is the default one provided by direnv that puts the layout dir at `$PWD/.direnv`. The symptom of this was

```
touch: cannot touch '/home/gmacon/.cache/direnv/layouts/.direnv/nix-profile-*.rc': No such file or directory
touch: cannot touch '/home/gmacon/.cache/direnv/layouts/.direnv/flake-profile-*.rc': No such file or directory
```

in the output and it not loading the environment correctly. Looking at the code, I think that's because it's also calling `direnv exec "/home/gmacon/.cache/direnv/layouts" true` and that directory naturally contains no `.envrc` to load.

ETA: for the immediate future, I've removed my custom `direnv_layout_dir` to get unblocked, but I'll try to implement a fix some evening this week, too.

bbenne10:
Closes #400.
 
 Cannot request @gmacon for review, but would appreciate his feedback.

### DIFF
--- a/direnvrc
+++ b/direnvrc
@@ -212,6 +212,7 @@ use_flake() {
 
   flake_expr="${1:-.}"
   flake_dir="${flake_expr%#*}"
+  flake_dir=${flake_dir#"path:"}
 
   if [[ "$flake_expr" = -* ]]; then
     if [[ -n "$2" ]]; then
--- a/direnvrc
+++ b/direnvrc
@@ -253,9 +253,9 @@ use_flake() {
   then
     if [[ "$_nix_direnv_manual_reload" == "1" && -z "${_nix_direnv_force_reload-}" ]]; then
       if [[ -e "$profile_rc" ]]; then
-        log_status "nix-direnv: cache is out of date. use \"nix-direnv-reload\" to reload"
+        log_error "nix-direnv: cache is out of date. use \"nix-direnv-reload\" to reload"
       else
-        log_status "nix-direnv: cache does not exist. use \"nix-direnv-reload\" to create it"
+        log_error "nix-direnv: cache does not exist. use \"nix-direnv-reload\" to create it"
       fi
 
     else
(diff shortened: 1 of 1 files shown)


direnvrc
```diff
@@ -34,23 +34,36 @@ _nix_direnv_preflight () {
...
+  mkdir -p "$layout_dir/bin"
+  # N.B. This script relies on variable expansion in *this* shell.
+  # (i.e. The written out file will have the variables expanded)
+  # If the source path changes, the script becomes broken.
+  # Because direnv_layout_dir is user controlled,
+  # we can't assume to be able to reverse it to get the source dir
+  # So there's little to be done about this.
+  cat > "${layout_dir}/bin/nix-direnv-reload" <<-EOF
+#!/usr/bin/env bash
+set -ex
```
gmacon, line 46: I'm assuming `set -x` was left over from debugging.

bbenne10: Done. Thanks.

bbenne10:
Realized that the earlier `mkdir -p` without a check was not gonna fly - it will print an error if the call fails.

gmacon (approved):
LGTM

direnvrc
```diff
@@ -37,16 +37,34 @@ _nix_direnv_preflight () {
...
+# rebuild the cache forcefully
+_nix_direnv_force_reload=1 direnv exec "$PWD" true
+
+# Update the mtime for .envrc.
+# This will cause direnv to reload again - but without re-building.
+touch "$PWD/.envrc"
+
+# Also update the timestamp of whatever profile_rc we have.
+# This makes sure that we know we are up to date.
+touch -r "$PWD/.envrc" ${layout_dir}/*.rc
```
Mic92, line 65: From my understanding layout_dir gets expanded, so we need to quote it in the script:

```suggestion
touch -r "$PWD/.envrc" "${layout_dir}"/*.rc
```

Mic92 (approved):

PR #407, opened 2023-11-17
Fix Nix file order docs

infinisil:
shell.nix file is used first if it exists, see https://github.com/nix-community/nix-direnv/blob/9178f71653285fff85b6a8712821b95273f95582/direnvrc#L340-L344

### DIFF
--- a/README.md
+++ b/README.md
@@ -303,8 +303,8 @@ when deciding to update its cache.
     * `.envrc`,
     * A single nix file. In order of preference:
         + The file argument to `use nix`
-        + `default.nix` if it exists
         + `shell.nix` if it exists
+        + `default.nix` if it exists
 
 - for `use flake` this is:
     * `~/.direnvrc`


infinisil:
Though ideally it would also use `default.nix` if it exists, a lot of `shell.nix`'s just defer to `default.nix`

Mic92:
Thanks!

Mic92:
> Though ideally it would also use `default.nix` if it exists, a lot of `shell.nix`'s just defer to `default.nix`

We need to stay compatible with `use nix` from the direnv stdlib instead. So this is not an option.

infinisil:
Ah, I guess yeah the default file to load should stay the same, but I think both files should be tracked for changes by default.

infinisil:
Opened https://github.com/nix-community/nix-direnv/issues/409 to track this further

Issue #409, opened 2023-11-18
Both `shell.nix` and `default.nix` should be watched for changes by default


infinisil:
A lot of `shell.nix` files just defer to `default.nix`. But currently only `shell.nix` is watched for changes by default, which means that changes to the `default.nix` won't update the environment.

So I propose to both watch `shell.nix` and `default.nix` by default.

Originally discussed in https://github.com/nix-community/nix-direnv/pull/407

PR #410, opened 2023-11-19
use_nix: watch for both shell.nix and default.nix

Mic92:
In many cases they will shell.nix will depend on default.nix and when a user switches from default.nix to shell.nix we also want to invalidate the cache.

fixes https://github.com/nix-community/nix-direnv/issues/409


bbenne10:
This looks fine to me.

Issue #408, opened 2023-11-17
`nix_direnv_watch_file` should be `watch_file`


infinisil:
If you want to make sure your environment is both automatically updated when files change, and updated to the correct value, and you don't know if the user has `nix-direnv` installed, you need to do something like this:

```bash
if has nix_direnv_watch_file; then
  nix_direnv_watch_file file.nix
else
  watch_file file.nix
fi

use nix
```

It would be much nicer if `nix-direnv` provided its own version of `watch_file` instead, such that a simple `watch_file file.nix` works.

`nix-direnv` would then also have the correct behavior by default for projects whose .envrc doesn't do anything special for nix-direnv.

Mic92:
What if make `nix_watches+=()` the api? This will work also if you don't have nix-direnv.

infinisil:
As far as I can see, there's never a case where you'd want to `watch_file` but not `nix_direnv_watch_file`, so overriding it would be fine.

Using `nix_watches+=()` would still require repos that were just considering `direnv` to make adjustments for `nix-direnv`.

Mic92:
@zimbatm I guess there is no other api to check if a file is beeing watched by direnv?

bbenne10:
@zimbatm has asked that we avoid monkey patching the direnv stdlib. 

However, I see no reason you couldn't do this:

```bash
if has nix_direnv_watch_file; then
  alias wf=nix_direnv_watch_file
else
  alias wf=watch_file
fi

wf flake.nix
```

infinisil:
The problem with any approach that doesn't just use `watch_file` is that it requires all projects using `direnv` to explicitly add support for `nix-direnv` as above, and otherwise their caching is just broken for all `nix-direnv` users.

bbenne10:
This is true. At work, we've taken the approach that using or not using `nix-direnv` is a *project* specification, which makes this a decision that we don't allow non-principal developers to override (Of course, they can discuss and influence direction, but we commit our `.envrc` to git to keep it consistent). We then use `source_url` to pin to a revision.

Yes - I agree that this is a bit frustrating. But the fact that both `direnv` and `nix-direnv` provide somewhat matching interfaces is a bit of an implementation detail in my eyes (and honestly - one I wish I could undo at this point. The confusion between the two projects has come up repeatedly in issues. I'd much prefer `use_flake_cache` or something...)

infinisil:
[...]

PR #414, opened 2023-11-23
use direnv-builtin `DIRENV_WATCHES` to list watched files

kingarrrt:
`nix_direnv_watch_file` is replaced with a function that wraps `watch_file` and issues a deprecation warning.

See #408 (which this fixes) for discussion.

### DIFF
--- a/direnvrc
+++ b/direnvrc
@@ -197,8 +197,39 @@ _nix_argsum_suffix() {
 }
 
 nix_direnv_watch_file() {
+  # shellcheck disable=2016
+  log_error '`nix_direnv_watch_file` is deprecated - use `watch_file`'
   watch_file "$@"
-  nix_watches+=("$@")
+}
+
+# set by nix build - see ./default.nix
+JQ=
+
+_jq() {
+  if [[ -n $JQ ]]
+  then
+    $JQ "$@"
+  elif has jq
+  then
+    jq "$@"
+  else
+    nix-shell --packages jq --run "$(join_args jq "$@")"
+  fi
+}
+
+_nix_direnv_watches() {
+  local -n _watches=$1
+  # DIRENV_WATCHES is json | gzip (without header and trailer) | base64url
+  # allow files under $XDG_DATA_HOME/direenv/allow are filtered out as not relevant
+  # gzip header: https://datatracker.ietf.org/doc/html/rfc1952#page-6
+  mapfile -td "" _watches < <(
+    # shellcheck disable=2016
+    tr -- "-_" "+/" <<< "$DIRENV_WATCHES" \
+      | base64 --decode \
+      | (echo -ne "\x1f\x8b\x08\x00\x00\x00\x00\x00"; cat /dev/stdin) \
+      | (gzip --decompress 2>/dev/null || true) \
+      | _jq --join-output '.[] | select(.Exists ) | .Path | select(. | test($ENV.XDG_DATA_HOME + "/direnv/allow") | not) | (. + "\u0000")'
+  )
 }
 
 _nix_direnv_manual_reload=0
@@ -229,7 +260,7 @@ use_flake() {
     files_to_watch+=("$flake_dir/flake.nix" "$flake_dir/flake.lock" "$flake_dir/devshell.toml")
   fi
 
-  nix_direnv_watch_file "${files_to_watch[@]}"
+  watch_file "${files_to_watch[@]}"
 
   local layout_dir profile
   layout_dir=$(direnv_layout_dir)
@@ -238,8 +269,10 @@ use_flake() {
   local flake_inputs="${layout_dir}/flake-inputs/"
 
   local need_update=0
+  local watches
+  _nix_direnv_watches watches
   local file=
-  for file in "${nix_watches[@]}"; do
+  for file in "${watches[@]}"; do
     if [[ "$file" -nt "$profile_rc" ]]; then
       need_update=1
       break
@@ -388,11 +421,13 @@ use_nix() {
     esac
   done
 
-  nix_direnv_watch_file "$HOME/.direnvrc" "$HOME/.config/direnv/direnvrc" ".envrc" "shell.nix" "default.nix"
+  watch_file "$HOME/.direnvrc" "$HOME/.config/direnv/direnvrc" ".envrc" "shell.nix" "default.nix"
 
   local need_update=0
+  local watches
+  _nix_direnv_watches watches
   local file=
-  for file in "${nix_watches[@]}"; do
+  for file in "${watches[@]}"; do
     if [[ "$file" -nt "$profile_rc" ]]; then
       need_update=1
       break
--- a/direnvrc
+++ b/direnvrc
@@ -139,7 +139,7 @@ _nix_import_env() {
 
   eval "$(< "$profile_rc")"
   # `nix print-dev-env` will create a temporary directory and use it as TMPDIR
-  # We cannot rely on this directory being availble at all times,
+  # We cannot rely on this directory being available at all times,
   # as it may be garbage collected.
   # Instead - just remove it immediately.
   # Use recursive & force as it may not be empty.
@@ -188,7 +188,7 @@ _nix_argsum_suffix() {
     elif has shasum; then
       out=$(shasum <<< "$1")
     else
-      # degrate gracefully both tools are not present
+      # degrade gracefully both tools are not present
       return
     fi
     read -r checksum _ <<< "$out"
--- a/direnvrc
+++ b/direnvrc
@@ -202,27 +202,21 @@ nix_direnv_watch_file() {
   watch_file "$@"
 }
 
-# set by nix build - see ./default.nix
-JQ=
-
-_jq() {
-  if [[ -n $JQ ]]; then
-    $JQ "$@"
-  elif has jq; then
-    jq "$@"
-  else
-    nix-shell --packages jq --run "$(join_args jq "$@")"
-  fi
-}
-
 _nix_direnv_watches() {
   local -n _watches=$1
-  # allow files under $XDG_DATA_HOME/direnv/allow are filtered out as not relevant
-  mapfile -td "" _watches < <(
-    # shellcheck disable=2016
-    direnv show_dump "${DIRENV_WATCHES}" \
-      | _jq --join-output '.[] | select(.Exists ) | .Path | select(. | test($ENV.XDG_DATA_HOME + "/direnv/allow") | not) | (. + "\u0000")'
-  )
+  while IFS= read -r line; do
+    local regex='"path": "(.+)"$'
+    if [[ "$line" =~ $regex ]]; then
+      local path="${BASH_REMATCH[1]}"
+      if [[ "$path" == "${XDG_DATA_HOME:-${HOME:-/var/empty}/.local/share}/direnv/allow/"* ]]; then
+        continue
+      fi
+      # expand new lines and other json escapes
+      # shellcheck disable=2059
+      path=$(printf "$path")
+      _watches+=("$path")
+    fi
+  done < <(direnv show_dump "${DIRENV_WATCHES}")
 }
 
 _nix_direnv_manual_reload=0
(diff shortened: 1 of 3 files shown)


bbenne10:
I will not personally merge this in the current state of affairs. I mentioned this on #408 as well, but to reiterate: @zimbatm has [*explicitly requested*](https://github.com/nix-community/nix-direnv/pull/148#discussion_r812668733) we not do this. If their opinion changes, I'll be fine to merge - until then, I respect them and their software too much to go against their wishes on this.

kingarrrt:
Somewhat contradictory... You respect them and their software, too much, but don't care for its implementation?

In any case, it is what it is, and I doubt they're gonna change it. For reference, the implementation is at https://github.com/direnv/direnv/blob/master/gzenv/gzenv.go.

Objectively, there's nothing very difficult about `gzip` and `base64url` - it's bog standard. You already depend on `coreutils` (for `base64`), this adds `gzip`, which is ubiquitous. It could be done without `jq`, but this would be more complex, requiring `sed` or `awk`.

The implementation here is a simple bash pipeline. It cannot be simpler. An equivalent Python one-liner is below, if you prefer, but this of course requires a Python interpreter, which I thought a bit heavy.

```bash
python -c "import base64, json, os, zlib; print('\n'.join([p['Path'] for p in json.loads(zlib.decompress(base64.urlsafe_b64decode(os.environ['DIRENV_WATCHES'])).decode()) if p['Exists'] and not p['Path'].startswith(os.environ['XDG_DATA_HOME'] + '/direnv/allow')]))"
```

(I should have commented in the patch. The `$XDG_DATA_HOME/direnv/allow` file is filtered from the list as not relevant.)

In any case, as covered in #408, `nix_direnv_watch_file` is a serious problem in that it introduces a hard dependency on this project.

You advertise as:

> A faster, persistent implementation of direnv's use_nix and use_flake, to replace the built-in one.

This is (currently) incorrect.

I appreciate your software and want to use it, but cannot as it stands. Do you agree that there is a problem?

bbenne10:
> Somewhat contradictory... You respect them and their software, too much, but don't care for its implementation?

I very often respect the net result of a developer's code, but disagree on implementation details. 

> In any case, it is what it is, and I doubt they're gonna change it. For reference, the implementation is at https://github.com/direnv/direnv/blob/master/gzenv/gzenv.go.

I've read it, thanks though.

> Objectively, there's nothing very difficult about `gzip` and `base64url` - it's bog standard. You already depend on `coreutils` (for `base64`), this adds `gzip`, which is ubiquitous. It could be done without `jq`, but this would be more complex, requiring `sed` or `awk`.

"Difficult" != "Complex"
This is a simple implementation, but an exposed list of files, separated by the record-separator character would be even easier and nearly as robust.

> In any case, as covered in #408, `nix_direnv_watch_file` is a serious problem in that it introduces a hard dependency on this project.
>
> You advertise as:
> 
> > A faster, persistent implementation of direnv's use_nix and use_flake, to replace the built-in one.
> 
> This is (currently) incorrect.
>
 
No - it does not add a hard dependency. #408 details how to work around the exposed interface inconsistencies. Alias `wf` to `nix-direnv-watch-file` or `watch-file` depending on the builtin `has` function's return value. Inconvenient? Yes - certainly. Not insurmountable when you have the entirety of bash at your fingertips.

> I appreciate your software and want to use it, but cannot as it stands. Do you agree that there is a problem?

As I made clear in my responses on #408 - I personally think that trying to match the names and interface exposed by the standard direnv implementation of `use nix` and `use flake` to be a bit of a mistake, but hindsight is 20/20 and we are where we are. Saying you "cannot use it" is a bit of a misdirect though - it is 100% possible to use `nix-direnv` in a scenario where you want to support direnv standard `use_{nix,flake}` also - you just have to do some papering over the interface differences.

Please do not take my initial reaction as indicative of some overall apathy in fixing this or some dislike of the solution - you did what you had to do to fix the problem in this software. My frustration stemmed from the fact that it appeared initially that you'd ignored my statements in #408 and opened a PR fixing the problem in the exact way I had said I didn't want to do it. This seems to do the right thing in light of the underlying implementation of `DIRENV_WATCHES`. I merely want to ruminate a bit on the implementation complexity vs gains in UX before I commit to merging this though.

kingarrrt:
Thank you for your considered reply.

Just to touch on the proposed workaround. It doesn't work for my (and others, @infinisil) use cases. "Hard dependency" was perhaps the wrong term, I can't justify extra noise in `.envrc` purely to support a tool that I want to use.

I believe that matching the direnv interface is a good thing, and that it would be to this project's advantage if it was a pure drop-in over direnv, requiring no boilerplate.

bbenne10:
I'd still like to see what @Mic92 has to say on the matter, but I think I'm okay with merging this. It does just enough to integrate with the direnv internals to alleviate the need for a wrapping function but not more than necessary. I haven't given thought to the exception you made for `$XDG_DATA_HOME/direnv/allow` and whether we need to add more exceptions though...

I have opened a discussion about what it would take to merge our version of `use_nix` and `use_flake` upstream on the direnv matrix channel (#direnv:numtide.com) and from what has been said so far, @zimbatm would prefer to divest of most (if not all) `use` and `layout` functions from direnv's core. This is interesting and I think I agree with that approach, as it democratizes the ability to get `use` or `layout` functions. (Python developers can work on `use python` and `layout python`; Ruby devs on the equivalent Ruby ones; Nix developers on the Nix ones; etc), but the question is how we'd LOAD the installed plugins. The discussion is ongoing, but I think that the takeaway so far is that once that were to happen - nix-direnv would have a bit of work to do to comply, but we would be the de facto standard for nix integration.

direnvrc
```diff
@@ -197,8 +197,19 @@ _nix_argsum_suffix() {
...
   watch_file "$@"
-  nix_watches+=("$@")
+}
+
+_nix_direnv_watches() {
+  # DIRENV_WATCHES is json | gzip (without header and trailer) | b64url
+  # gzip header: https://datatracker.ietf.org/doc/html/rfc1952#page-6
+  tr -- "-_" "+/" <<< "$DIRENV_WATCHES" \
+    | base64 --decode \
+    | (echo -ne "\x1f\x8b\x08\x00\x00\x00\x00\x00"; cat /dev/stdin) \
```
Mic92, line 210: It's cool to re-use direnv's watches here, I am only a bit on the fence with depending on jq. Nix is often introduced in new teams through direnv and those folks might not have all dependencies installed, which can cause friction. I wonder if we can make somehow a poly fill that uses nix for json parsing as a fallback?

```suggestion
    | (echo -ne "\x1f\x8b\x08\x00\x00\x00\x00\x00"; base64 --decode) \
```

Mic92: Also do we have base64 on macOS? We might have to depend on openssl base64?

direnvrc
```diff
@@ -197,8 +197,19 @@ _nix_argsum_suffix() {
...
+    | (gzip --decompress 2>/dev/null || true) \
```
Mic92, line 211: Same question for gzip.

direnvrc
kingarrrt: pushed with polyfill, macOS has base64

direnvrc
kingarrrt: macOS has gzip

kingarrrt:
I've pushed an updated version with a polyfill using `nix-shell` for `jq`. It is of course slow. Do you think it might be worth issuing a warning if it falls back to this?

I also changed the parsed watches to be NUL-separated which should be more robust.

infinisil:
Is it really better to do this instead of just overriding `watch_file`? This feels very brittle to me

Mic92:
> Is it really better to do this instead of just overriding `watch_file`? This feels very brittle to me

We don't want to override more builtins. You can also see the complexity that comes from overriding just two builtins.

Mic92:
@kingarrrt I also added the functionality to direnv itself: https://github.com/direnv/direnv/pull/1198

direnvrc
```diff
@@ -197,8 +197,39 @@ _nix_argsum_suffix() {
...
+  # DIRENV_WATCHES is json | gzip (without header and trailer) | base64url
+  # allow files under $XDG_DATA_HOME/direenv/allow are filtered out as not relevant
+  # gzip header: https://datatracker.ietf.org/doc/html/rfc1952#page-6
+  mapfile -td "" _watches < <(
+    # shellcheck disable=2016
+    tr -- "-_" "+/" <<< "$DIRENV_WATCHES" \
+      | base64 --decode \
+      | (echo -ne "\x1f\x8b\x08\x00\x00\x00\x00\x00"; cat /dev/stdin) \
+      | (gzip --decompress 2>/dev/null || true) \
+      | _jq --join-output '.[] | select(.Exists ) | .Path | select(. | test($ENV.XDG_DATA_HOME + "/direnv/allow") | not) | (. + "\u0000")'
```
Mic92, line 231: We can replace most of this with:

```
direnv show_dump "${DIRENV_WATCHES}"
```

Than we are only left with `jq`

direnvrc
```diff
@@ -197,8 +197,39 @@ _nix_argsum_suffix() {
...
+
+_jq() {
+  if [[ -n $JQ ]]
+  then
+    $JQ "$@"
+  elif has jq
+  then
+    jq "$@"
+  else
+    nix-shell --packages jq --run "$(join_args jq "$@")"
```
Mic92, line 216: It makes we wonder if we could just use `nix eval` and parse the json... it probably has better performance than evaluating and downloading `jq`.

direnvrc
```diff
@@ -197,16 +197,33 @@ _nix_argsum_suffix() {
...
 nix_direnv_watch_file() {
+  # shellcheck disable=2016
+  log_error '`nix_direnv_watch_file` is deprecated - use `watch_file`'
   watch_file "$@"
-  nix_watches+=("$@")
+}
+
+_nix_direnv_watches() {
+  local -n _watches=$1
+  while IFS= read -r line; do
```
Mic92, line 207: We can drop this code when we get https://github.com/direnv/direnv/pull/1198

Mic92:
Ok. After my refactoring this touches a lot less builtins in direnv. And if we get the PR merged it will be even less. Also now it no longer adds any new dependencies.

Mic92:
> I'd still like to see what @Mic92 has to say on the matter, but I think I'm okay with merging this. It does just enough to integrate with the direnv internals to alleviate the need for a wrapping function but not more than necessary. I haven't given thought to the exception you made for `$XDG_DATA_HOME/direnv/allow` and whether we need to add more exceptions though...
> 
> I have opened a discussion about what it would take to merge our version of `use_nix` and `use_flake` upstream on the direnv matrix channel (`#direnv:numtide.com`) and from what has been said so far, @zimbatm would prefer to divest of most (if not all) `use` and `layout` functions from direnv's core. This is interesting and I think I agree with that approach, as it democratizes the ability to get `use` or `layout` functions. (Python developers can work on `use python` and `layout python`; Ruby devs on the equivalent Ruby ones; Nix developers on the Nix ones; etc), but the question is how we'd LOAD the installed plugins. The discussion is ongoing, but I think that the takeaway so far is that once that were to happen - nix-direnv would have a bit of work to do to comply, but we would be the de facto standard for nix integration.

I am aware of these plans, but to me direnv's stdlib is also one of the biggest strengths. If someone has to check 5 different places to get high-quality plugins for each ecosystem (the nodejs community would probably end up with 3 for the same thing) than this is worse user experience. I trust zimbatm to be more thoughtful when it comes to backwards compatibility compared to many other project maintainer.

bbenne10:
> I am aware of these plans, but to me direnv's stdlib is also one of the biggest strengths. If someone has to check 5 different places to get high-quality plugins for each ecosystem (the nodejs community would probably end up with 3 for the same thing) than this is worse user experience. I trust zimbatm to be more thoughtful when it comes to backwards compatibility compared to many other project maintainers.

IMHO if the nodejs community shoots themselves in the foot with this change, I'm not sure I think it is direnv's problem in the first place. There are things I think we would need to implement (naming conventions, a plugin repository with a basic search feature or some other way to find plugins that've been "blessed") and we'd need to find maintainers for the separate implementations (or maybe only for the "sufficiently complex" ones), but I think that the separation of implementation out into smaller chunks might help improve the contributor base to each specialized bit. It's certainly more managerial work to make everything work together, but it makes the technical bits more approachable imo.

PR #415, opened 2023-11-26
drop dependency on gnugrep

Mic92:


bbenne10 (approved):
Using BASH_REMATCH is a good solution here. LGTM!

Mic92:
@mergify queue

Mic92:
@mergify refresh

PR #417, opened 2023-11-26
_nix_direnv_watches: update path regex to match capitalized path

kingarrrt:
Small problem with a20b32d. In my environment (direnv 2.32.2) the keys in `DIRENV_WATCHES` are capitalized.

### DIFF
--- a/direnvrc
+++ b/direnvrc
@@ -208,7 +208,7 @@ _nix_direnv_watches() {
     return
   fi
   while IFS= read -r line; do
-    local regex='"path": "(.+)"$'
+    local regex='"[Pp]ath": "(.+)"$'
     if [[ "$line" =~ $regex ]]; then
       local path="${BASH_REMATCH[1]}"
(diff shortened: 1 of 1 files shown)


bbenne10:
...I cannot reproduce this. 

You've fixed this in a backwards compatible way, so I'm happy to merge - but I wonder now why I can't reproduce. The version of direnv I have is only a point release higher (2.32.3), so I wouldn't expect something like this to change?

A search of the direnv's issue and pr trackers yield very little. I wonder what the causing this to happen?

bbenne10:
Ok. [This](https://github.com/direnv/direnv/pull/1086) is what does the changing of the serialization format and you'll note that that is included in the that point release. To support everything except the most recent point release, this is 100% a needed change.

kingarrrt:
Well spotted. I couldn't work it out.

bbenne10:
Simply knowing how to use the blame feature in the UI coupled with stumbling onto the right struct in the direnv code. :P

PR #419, opened 2023-11-27
Release 3.0.0

bbenne10:
We just released 2.5.0 yesterday, but it strikes me that removing API surface is actually a major version bump. I'm opening a PR to handle the release to discuss a major release (wouldn't have bothered for a minor or point release)

Yes, we have back compat but we should still bump the major and keep the back compat until next major release.

Note that I didn't use the existing scripts, so a tag will need to be created after merge (or some other inclusion). 
The sha256 for fetchurl should be correct, however.


bbenne10:
That's just fine with me.

Issue #412, opened 2023-11-21
fallback to previous working env if building new env with changes fails


robert-elles:
Hi,

when I make changes to my environment and they fail to build direnv just looses the whole environment. This is often unhelpful for the ongoing development work to fix the changes. It would be better if the previous environment stays active and can also be activated in a new terminal until the switch can be made.

Thank you and kind regards

robert-elles:
FYI: I am using flakes

bbenne10:
I think that #420 should resolve this. Can you test with latest master and report back if the experience is what you want and expect? Thanks!

PR #420, opened 2023-11-27
Delay invalidating cache and gcroots until print-dev-env call succeeds

bbenne10:
We currently eagerly invalidate the gcroots and old profile rc, assuming that the devshell is in a usable state.
If this assumption does not hold,
we can invalidate a working state for a broken one.

Here we just delay calling `_nix_clean_old_gcroots`
until we know that we're in a usable state.

In the case that the flake is in an unusable state, this simply reuses the latest working state.

This should address #412.


bbenne10:
I haven't looked into test failures, but will when I find a moment.

Mic92:
@mergify queue

Mic92:
@bbenne10 I get the same errors locally. Presumably only present after doing the rebase?

bbenne10:
I am starting to believe that there's something going wrong with the interaction between the creation of the `tmp_profile_rc` and the `_nix_add_gcroot` call which causes `direnv exec` to exit non-zero. I haven't personally modified this behavior in this MR so I think we've got a longer standing bug, maybe?

I'm looking into it a bit more deeply now.

bbenne10:
Of course, I post and then figure out the bug. 

`_nix_clean_old_gcroots` removes `$tmp_profile`. Moving it where I did causes it to remove `$tmp_profile` in between populating `$tmp_profile_rc` and the subsequent call to `nix build` (as part of `_nix_add_gcroot`), which causes the call to `nix build` to fail and thus the `direnv exec` call exits 1. I'll poke around to see how I can resolve this.

bbenne10:
There we go. Tests are passing now.

I have rebased on #427 to keep in line with the logic and logging changes in there. We should get that merged first.

Mic92:
@mergify queue

PR #422, opened 2023-11-28
refactor tests and dev env

kingarrrt:
This attempts to extend and clean up the test suite and dev environment. Not complete but a good start (I think) and I hope not controversial. I was surprised that you are not dogfooding with an `.envrc`, and have added that in.

The tests are now self-contained, so unaffected by external state, and equally they won't mess with your `$HOME`. The bear in the corner remains - gc tests still use the real store, which is painful.

One thing this flags up is your minimum nix requirement. Nix 2.4 was removed from Nixpkgs in July/22. The lowest version you can support in Nixpkgs is 2.10, but this has it's own problems, see the last commit. Maybe time to bump the requirement?


bbenne10:
Hah. I was just thinking that I would like to get the tests working more transparently (and figure out why they aren't running on #420?). I'll look this over soon.

kingarrrt:
Er, I've got two more branches, https://github.com/kingarrrt/nix-direnv/tree/cleanup and https://github.com/kingarrrt/nix-direnv/tree/cached. The first is a general cleanup, the second re-implements the cache keyed on the hash of watched files. Unfortunately each builds on the last, including this one, and I can't create pull requests without a bunch of duplicate commits. I thought Github would deal with the situation but no... I'm not quite sure what to do.

Re the failures: I hadn't realized Garnix ran `flake check`. I've got to head out but will come back later and move these tests to be exposed under something that won't get picked by it, since the tests are still impure so cannot work in this context.

tests/test_use_flake.py
```diff
@@ -0,0 +1,37 @@
+from __future__ import annotations
```
Mic92, line 1: Do we still need this? We don't need to support python 3.7 anymore for our internal test suite.

pyproject.toml
```diff
@@ -1,20 +1,58 @@
...
+parallel = true
+source = ["."]
+relative_files = true
 
-select = ["E", "F", "I"]
-ignore = [ "E501" ]
+[tool.coverage.html]
+show_contexts = true
 
+[tool.coverage.report]
```
Mic92, line 10: Only our test code is using python, so I don't think we need code coverage for this.

default.nix
```diff
@@ -1,22 +1,59 @@
...
+      touch $out
+    '';
+    test = finalAttrs.overrideAttrs {
+      name = "test-${finalAttrs.name}-${nix.name}";
+      nativeBuildInputs = [
+        direnv
+        nix
+      ] ++ (with python3.pkgs; [
+        pytestCheckHook
+        pytest-cov
```
Mic92, line 46: So we might not need this?

default.nix
```diff
@@ -1,22 +1,59 @@
...
+        pytest-randomly
+        pytest-sugar
```
Mic92, line 48: Checked the number of dependencies and it seems reasonable.

Mic92:
There are certainly some good things in this PR but I also would like to take this in a smaller scoped pull requests.

kingarrrt:
Understood. I'll split this down and resubmit.

PR #423, opened 2023-11-29
Prefer using `local` over `local -n`

ipetkov:
* zsh on macOS (the default shell) apparently does not support `local -n` and will error out if used

### DIFF
--- a/direnvrc
+++ b/direnvrc
@@ -204,7 +204,7 @@ nix_direnv_watch_file() {
 }
 
 _nix_direnv_watches() {
-  local -n _watches=$1
+  local _watches=$1
   if [[ -z "${DIRENV_WATCHES-}" ]]; then
     return
   fi


kingarrrt:
This would break it. `local -n` makes a named reference used by the caller.

bbenne10:
Using nix-direnv directly in zsh is not supported. I'm going to close this, but I think the discussion is worth continuing even so (namely "How did this get sourced in zsh?" and "If this is because of confusion over installation procedures, can we clarify installation procedures at all to help with that?"

ipetkov:
More context: I can reproduce this with direnv installed via homebrew

bbenne10:
This is explicitly noted in the README already.  We ask users to install a modern bash version into their profile. You can do this with homebrew or nix or any other packaging tool for macOS. If you're installing direnv from homebrew, is it not reasonable to also install bash from there? The version shipped with macOS is *ancient* and I'm afraid that this is not the only thing we'll run afoul of on a version that old.

I'll reopen if there's a change to the source branch that fixes for 3.2 without a bunch more complexity for modern bash versions, but I'm hesitant to say that we'll continue to support ancient bash versions on proprietary systems (even as a user of macOS + nix-direnv myself!)

Mic92:
We heavily depend on bash's regex feature. Is this even part of 3.2?

ipetkov:
Thanks for the context here! A bunch of colleagues hit this at work and I was hopeful there could be an easy solution but looks like there isn't. I'll suggest to them to install a newer version of bash

PR #425, opened 2023-11-29
More ruff lints

Mic92:


bbenne10:
Didn't get around to reviewing this before merge (it was about 3AM local time when you opened it and 8AM local time when you merged it, so I give myself a pass :P)

But I generally agree with the changes - though I think I'd like to see some of the ruff ignore lines removed and fixed in the medium term - we can get to that eventually though. No big rush.

PR #429, opened 2023-11-30
simplify required version checks and bring back support for bash 3

kingarrrt:
The =~ operator is bash 3, the only bash 4 feature used was named references introduced in 6997988. This changes that to use variable inheritance.

Tested on NixOS and macOS with direnv via both homebrew and nix.

Refs #423.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -22,11 +22,11 @@
     },
     "nixpkgs": {
       "locked": {
-        "lastModified": 1700856099,
-        "narHash": "sha256-RnEA7iJ36Ay9jI0WwP+/y4zjEhmeN6Cjs9VOFBH7eVQ=",
+        "lastModified": 1701174899,
+        "narHash": "sha256-1W+FMe8mWsJKXoBc+QgKmEeRj33kTFnPq7XCjU+bfnA=",
         "owner": "NixOS",
         "repo": "nixpkgs",
-        "rev": "0bd59c54ef06bc34eca01e37d689f5e46b3fe2f1",
+        "rev": "010c7296f3b19a58b206fdf7d68d75a5b0a09e9e",
         "type": "github"
       },
       "original": {
--- a/flake.nix
+++ b/flake.nix
@@ -6,9 +6,6 @@
   inputs.treefmt-nix.url = "github:numtide/treefmt-nix";
   inputs.treefmt-nix.inputs.nixpkgs.follows = "nixpkgs";
 
-  nixConfig.extra-substituters = [ "https://cache.garnix.io" ];
-  nixConfig.extra-trusted-public-keys = [ "cache.garnix.io:CTFPyKSLcx5RMJKfLo5EEPUObbA78b0YQ2DTCJXqr9g=" ];
-
   outputs = inputs @ { flake-parts, ... }:
     flake-parts.lib.mkFlake { inherit inputs; }
       ({ lib, ... }: {
--- a/templates/flake/flake.nix
+++ b/templates/flake/flake.nix
@@ -3,7 +3,7 @@
   inputs.nixpkgs.url = "github:NixOS/nixpkgs/nixpkgs-unstable";
   inputs.flake-utils.url = "github:numtide/flake-utils";
 
-  outputs = { nixpkgs, flake-utils }:
+  outputs = { nixpkgs, flake-utils, ... }:
     flake-utils.lib.eachDefaultSystem (system:
       let
         pkgs = nixpkgs.legacyPackages.${system};


bbenne10:
I've given this some thought and I think that "whatever version of bash is in nixpkgs stable" is the version we support. This seems the only reasonable way to ensure that everyone has the same version and reduces the burden on us to only check "does this work in nixpkgs stable". 

I know that's not *great* news for macOS users, but the situation isn't any worse than the one we have implicitly now and the burden of checking compatibility is greatly reduced. Does that sound reasonable? @Mic92: Do you have thoughts?

direnvrc
```diff
@@ -31,28 +40,38 @@ _nix() {
   nix --extra-experimental-features "nix-command flakes" "$@"
 }
 
+_require_version() {
+  local cmd=$1 version=$2 required=$3
+  if ! printf "%s\n" "$required" "$version" | sort --check=quiet --version-sort; then
```
Mic92, line 45: Does this check works with the macos sort? https://ss64.com/osx/sort.html

Mic92: Otherwise we could also try to detect if the bash version supports regex instead?

Mic92: Otherwise here is a pure bash version:

```
version_compare () {
    if [[ $1 == $2 ]]
    then
        return 0
    fi
    local IFS=.
    local i ver1=($1) ver2=($2)
    # fill empty fields in ver1 with zeros
    for ((i=${#ver1[@]}; i<${#ver2[@]}; i++))
    do
        ver1[i]=0
    done
    for ((i=0; i<${#ver1[@]}; i++))
    do
        if [[ -z ${ver2[i]} ]]
        then
            # fill empty fields in ver2 with zeros
            ver2[i]=0
        fi
        if ((10#${ver1[i]} > 10#${ver2[i]}))
        then
            return 1
        fi
        if ((10#${ver1[i]} < 10#${ver2[i]}))
        then
            return 2
        fi
    done
    return 0
}
```

Mic92:
> I've given this some thought and I think that "whatever version of bash is in nixpkgs stable" is the version we support. This seems the only reasonable way to ensure that everyone has the same version and reduces the burden on us to only check "does this work in nixpkgs stable".
> 
> I know that's not _great_ news for macOS users, but the situation isn't any worse than the one we have implicitly now and the burden of checking compatibility is greatly reduced. Does that sound reasonable? @Mic92: Do you have thoughts?

Mhm. I think having a version check that gives proper errors is also nice, because people can just slap in the `source_url` section and than it's being automatically used by every user of the project. And it's hard to control how people are going to setup their system. Direnv is often the gateway drug to introduce nix in teams and you don't want to cause there more friction than required.

direnvrc
kingarrrt: Verified on macOS 14.1.2

kingarrrt:
Pushed again with BASH_MIN_VERSION=5.2.

kingarrrt:
> Is there a possibility to add bash to the dependency list of nix's direnv derivation and homebrew's direnv package so that it picks a modern bash (which seems useful more generally for direnv) or something similar? It seems useful to know what runtime you can be expected to execute on with some certainty...

Nixpkgs already has the bash dependency: https://github.com/NixOS/nixpkgs/blob/master/pkgs/tools/misc/direnv/default.nix#L23

Homebrew needs it: https://github.com/Homebrew/homebrew-core/blob/master/Formula/d/direnv.rb

I'll try and fiddle a homebrew pull request later on.

I'll try to fiddle a

direnvrc
Mic92: Ok. Just to double check, you were using the builtin sort executable and not sort from coreutils here?

Mic92:
> Mhm. I think having a version check that gives proper errors is also nice, because people can just slap in the source_url section and than it's being automatically used by every user of the project. And it's hard to control how people are going to setup their system. Direnv is often the gateway drug to introduce nix in teams and you don't want to cause there more friction than required. It is a hard call for me, honestly. If we decide to explicitly say we support any given version, we MUST test against that version, I believe. We are making a commitment to that and should ensure we are holding ourselves to our promise and I am frankly not entirely sure how to do that. I COMPLETELY see that direnv is an entry point to the larger nix ecosystem and I want us to be a good ambassador. This seems extreme, but... Is there a possibility to add bash to the dependency list of nix's direnv derivation and homebrew's direnv package so that it picks a modern bash (which seems useful more generally for direnv) or something similar? It seems useful to know what runtime you can be expected to execute on with some certainty...

I think we can balance that based on how much ugly the required code looks like. Just now it looks like the more portable code is even shorter than what we currently have. We can revisit the decision later again if things are changing.

direnvrc
kingarrrt:
```
arthur@Mac-mini nix-direnv % command -v sort
/usr/bin/sort
arthur@Mac-mini nix-direnv % sort --version
2.3-Apple (165.0.4)
```

Mic92:
@mergify queue

kingarrrt:
homebrew bash dependency: https://github.com/Homebrew/homebrew-core/pull/156230

bbenne10:
> I think we can balance that based on how much ugly the required code looks like. Just now it looks like the more portable code is even shorter than what we currently have. We can revisit the decision later again if things are changing.

Completely agree that this is better and significantly simpler than our old approach. I do not think we should've avoided merging the major parts of this changeset, but rather I worry about *promising* backwards compatibility as far as we are with the merge of this PR. We are now saying with some certainty that we *know* we execute on bash 3.2, which I know I cannot keep up with without additional burden - either by researching on what version specific features were introduced into or explicit testing on 3.2, which isn't available in nixpkgs and so means I can only really do with changes to my profile on my macOS machine (a machine I would prefer to leave behind, to be perfectly honest). The burden of work on introducing new non-trivial changes is thus *already* increased for myself and for additional contributors. 

We can, for instance, introduce an explicit test against bash 3.2, but I am afraid (as with the `sort` compatibility question earlier in this PR) that we won't know we've introduced other breaking changes with that too. 

Again: I'm trying to find a balance between end-user utility and the overall burden on developers and contributors, not trying to stand in the way of good forward progress. I appreciate you all listening to my concerns. Thank you!

kingarrrt:
Yes. I'll see what I can do. The test is only going to verify that it fails the version check though, so nothing much.

@bbenne10 Note that 31984f6 makes no backwards compatibility promises.

bbenne10:
> @bbenne10 Note that [31984f6](https://github.com/nix-community/nix-direnv/commit/31984f60833b3a92a9a9f928691f1d8f6ea58839) makes no backwards compatibility promises.

Whoops. Somehow I missed this entirely. Sorry for the noise, but I'm happy to have an opportunity to spell out my thoughts on backwards compatibility and supported underlying software revisions.

PR #433, opened 2023-12-03
 use the new watch-print command of direnv 

Mic92:
This is only part of direnv 2.33.0, so we need to bump the minimum.
We probably wait to merge this PR until this direnv version is reasonably
widely available: https://packages.debian.org/de/sid/direnv


bbenne10:
I like these changes, but agree that we have to wait for a bit for full dispersal of the new direnv version. We should be able to keep an eye [here](https://repology.org/project/direnv/versions) to check on that.

bbenne10:
Debian 13 "Trixie" still has 2.32.1. We're still waiting for this one, unfortunately.

PR #438, opened 2023-12-09
deprecate nix_direnv_watch_file

Mic92:


Mic92:
After that I would release nix-direnv 3.0.0

bbenne10 (approved):
Let's do it. Happy to get 3.0 dropped.

PR #441, opened 2023-12-14
make self-contained

### LINKED ISSUE #440: nix-direnv fails to correctly recognize bash version
Hello,

After upgrading my system, I get the following error when entering a previously functioning environment

```
direnv: loading ~/Repositories/myfolder
direnv: using flake
sort: Unknown option 'check=quiet' (see "sort --help")
direnv: nix-direnv: minimum required bash version is 5.2 (installed: 5.2.21(1)-release)
direnv: error exit status 1
```

This is  a valid version, but direnv fails to parse it as valid

The code responsible for this is :

```bash
_require_version() {
  local cmd=$1 version=$2 required=$3
  if ! printf "%s\n" "$required" "$version" | sort --check=quiet --version-sort; then
    _nix_direnv_fatal "minimum required $cmd version is $required (installed: $version)"
  fi
}

```
For some reason `sort` doesn't recognize the full form options (`--check=quiet --version-sort`), using the short form options (`-C -V`) does not fix the problem

If I look into using sort to sort `5.2` and `5.2.21(1)-release` it seems like the sorting is correct ? I don't understand how this error happens

Best,

kingarrrt:
This prepends `coreutils` and `direnv` to `$PATH` in addition to `nix` in order to make this self-contained.

Fixes #440.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -1,17 +1,47 @@
-{ stdenv, nix, lib }:
+{ resholve, lib, coreutils, direnv, nix }:
 
-stdenv.mkDerivation {
-  name = "nix-direnv";
+# resholve does not yet support `finalAttrs` call pattern hence `rec`
+# https://github.com/abathur/resholve/issues/107
+resholve.mkDerivation rec {
+  pname = "nix-direnv";
+  version = "3.0.0";
 
-  src = ./.;
-
-  postPatch = ''
-    sed -i "2iNIX_BIN_PREFIX=${nix}/bin" direnvrc
-  '';
+  src = builtins.path {
+    path = ./.;
+    name = pname;
+  };
 
   installPhase = ''
-    install -m400 -D direnvrc $out/share/nix-direnv/direnvrc
+    install -m400 -D direnvrc $out/share/${pname}/direnvrc
   '';
+
+  solutions = {
+    default = {
+      scripts = [ "share/${pname}/direnvrc" ];
+      interpreter = "none";
+      inputs = [ coreutils direnv nix ];
+      fake = {
+        builtin = [
+          "PATH_add"
+          "direnv_layout_dir"
+          "has"
+          "log_error"
+          "log_status"
+          "watch_file"
+        ];
+        function = [
+          # not really a function - this is in an else branch for macOS/homebrew that
+          # cannot be reached when built with nix
+          "shasum"
+        ];
+      };
+      keep."$cmd" = true;
+      execer = [
+        "cannot:${direnv}/bin/direnv"
+        "cannot:${nix}/bin/nix"
+      ];
+    };
+  };
 
   meta = with lib; {
     description = "A fast, persistent use_nix implementation for direnv";
--- a/default.nix
+++ b/default.nix
@@ -11,6 +11,11 @@ resholve.mkDerivation rec {
     name = pname;
   };
 
+  # drop min version checks which are redundant when built with nix
+  postPatch = ''
+    sed -i "/_require_version bash/,+2d" direnvrc
+  '';
+
   installPhase = ''
     install -m400 -D direnvrc $out/share/${pname}/direnvrc
   '';
--- a/direnvrc
+++ b/direnvrc
@@ -59,10 +59,6 @@ _nix_direnv_preflight() {
     _nix_direnv_fatal '$direnv environment variable was not defined. Was this script run inside direnv?'
   fi
 
-  if [[ -n ${NIX_BIN_PREFIX:-} ]]; then
-    PATH_add "$NIX_BIN_PREFIX"
-  fi
-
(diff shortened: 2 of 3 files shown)


Mic92:
Not quite sure this fully fixes the problem because what happens if people install nix-direnv not via nix but i.e. use source_url.

bbenne10:
Is it worth considering removing support for installing outside of nix workflows? Installing via nix would expand our possibilities (i.e. we could use jq and this problem would be solved via a simple anti-quotation). The Installation barrier would be higher. We could offset that by pointing at the flake so that they could override the nixpkgs' version with arbitrary commits. 

Everyone consuming this should have Nix. Nix now has support for flakes everywhere (2.10 being the earliest you can get in nixpkgs). And the whole purpose is to make using flakes/shell.nix easier in a project, so I think it reasonable to say that you must have a devShell set up for this project to use nix-direnv.

On the other hand - I can see a circular dependency happening whereby nix-direnv is installed as *part* of a devShell and so won't be on PATH until the devShell is loaded.

kingarrrt:
I was thinking only of the case where the install is via nix.

To take this a bit further, we could check if `sort` is a symlink to `coreutils` and if not install it, but this seems a lot of effort for an edge case.

default.nix
```diff
@@ -1,12 +1,12 @@
...
+{ stdenv, nix, lib, coreutils, direnv }:
 
 stdenv.mkDerivation {
   name = "nix-direnv";
 
   src = ./.;
 
   postPatch = ''
-    sed -i "2iNIX_BIN_PREFIX=${nix}/bin" direnvrc
+    sed -i '2iBIN_PREFIX="${nix}/bin ${coreutils}/bin ${direnv}/bin"' direnvrc
```
Mic92, line 9: Since we don't guard against environment variables from the outside, I would rename this to `NIX_DIRENV_BIN`

Mic92:
```suggestion
    sed -i '2iNIX_DIRENV_BIN="${nix}/bin ${coreutils}/bin ${direnv}/bin"' direnvrc
```

direnvrc
```diff
@@ -58,8 +58,9 @@ _nix_direnv_preflight () {
     _nix_direnv_fatal "\$direnv environment variable was not defined. Was this script run inside direnv?"
   fi
 
-  if [[ -n ${NIX_BIN_PREFIX:-} ]]; then
-    PATH_add "$NIX_BIN_PREFIX"
+  if [[ -n "${BIN_PREFIX:-}" ]]; then
```
Mic92, line 61:
```suggestion
  if [[ -n "${NIX_DIRENV_BIN:-}" ]]; then
```

direnvrc
```diff
@@ -58,8 +58,9 @@ _nix_direnv_preflight () {
...
+    # shellcheck disable=2086
+    PATH_add $BIN_PREFIX
```
Mic92, line 63:
```suggestion
    PATH_add $NIX_DIRENV_BIN
```

kingarrrt:
Just discovered https://github.com/abathur/resholve which deals with this in a much nicer way.

Lyndeno:
Is there a way to check what version of nix-direnv I am using? I updated to the latest nixpkgs which should be at 3.0.1 with this fix but I want to make sure. I am still running into the error I described in https://github.com/nix-community/nix-direnv/issues/448#issuecomment-1865831469.

Mic92:
nix_direnv_version

Lyndeno:
Here is my output:
```console
$ direnv allow
direnv: loading ~/Projects/nix-config/.envrc
3.0.1
direnv: using flake
sort: unrecognized option '--check=quiet'
BusyBox v1.36.1 () multi-call binary.

Usage: sort [-nrughMcszbdfiokt] [-o FILE] [-k START[.OFS][OPTS][,END[.OFS][OPTS]] [-t CHAR] [FILE]...

Sort lines of text

	-o FILE	Output to FILE
	-c	Check whether input is sorted
	-b	Ignore leading blanks
	-f	Ignore case
	-i	Ignore unprintable characters
	-d	Dictionary order (blank or alphanumeric only)
	-n	Sort numbers
	-g	General numerical sort
	-h	Sort human readable numbers (2K 1G)
	-M	Sort month
	-V	Sort version
	-t CHAR	Field separator
	-k N[,M] Sort by Nth field
	-r	Reverse sort order
	-s	Stable (don't sort ties alphabetically)
	-u	Suppress duplicate lines
	-z	NUL terminated input and output
direnv: nix-direnv: minimum required bash version is 5.2 (installed: 5.2.21(1)-release)
direnv: error exit status 1
```

using NixOS `24.05.20231227.cfc3698` using the `nix-direnv` option in home-manager

Mic92:
You are using an old version of nix-direnv. The current version doesn't have this flag.

PR #443, opened 2023-12-14
treefmt: use shfmt and statix

kingarrrt:
I propose enabling `shfmt` and `statix` in `treefmt`.

In addition this simplifies `treefmt.nix` and adds `direnvrc` to `treefmt` shell includes (default is `*.sh`).


Mic92:
@mergify queue

Mic92:
@mergify refresh

PR #444, opened 2023-12-14
envrc: watch direnvrc and *.nix

kingarrrt:
I believe this way catches all dependencies of the shell environment.


Mic92:
@mergify queue

PR #449, opened 2023-12-19
Fixup `$direnv` usage

kingarrrt:
This extends #447 by using `$direnv` to call direnv.

I've also modified the version check skip for nix to be less fragile.


Mic92:
@mergify queue

Issue #448, opened 2023-12-19
nix-direnv: minimum required bash version is 5.2 


francis2tm:
Hello,
I'm on Ubuntu 22.04

## Error

```
direnv: nix-direnv: minimum required bash version is 5.2 (installed: 5.1.16(1)-release)
```
And unfortunately, I cannot use nix-direnv since ubuntu's newest bash version is still 5.1.16. (and forcing a new version will probably kill my system)
Tried to use zsh instead, but apparently nix-direnv does not support any other shell than bash.

Any workaround that allows nix-direnv to become more flexible with the bash version?

Thanks

bbenne10:
Install bash from nixpkgs into your profile. You can do this in any number of ways (`nix-env`, home-manager, etc)

Mic92:
@kingarrrt do we actually need bash 5.2? Ubuntu 22.04 doesn't sound that old to me.

kingarrrt:
> @kingarrrt do we actually need bash 5.2? Ubuntu 22.04 doesn't sound that old to me.

We don't. Per https://github.com/nix-community/nix-direnv/pull/429#issuecomment-1834305543 it could be bash 4. We chose bash 5.2 based on what's available in nixpkgs and homebrew.

@francis2tm Can you explain how you installed direnv and how you are using nix-direnv?

Mic92:
Ok. I think we maybe should set it back to bash 4 than. Those versions are still in the latest redhat/centos versions AFAIK.

francis2tm:
> > @kingarrrt do we actually need bash 5.2? Ubuntu 22.04 doesn't sound that old to me.
> 
> We don't. Per [#429 (comment)](https://github.com/nix-community/nix-direnv/pull/429#issuecomment-1834305543) it could be bash 4. We chose bash 5.2 based on what's available in nixpkgs and homebrew.
> 
> @francis2tm Can you explain how you installed direnv and how you are using nix-direnv?

Of course! Essentially my objective is to enable nix flake with vscode, specially extensions (e.g. I want rust-analyzer extension to use flake's rust toolchain).
So, after a lot of research, it seems direnv can kickstart nix flake using a mix of direnv + a [direnv extension](https://marketplace.visualstudio.com/items?itemName=Rubymaniac.vscode-direnv) + nix-direnv

I installed direnv using ubuntu's package manager:
```
sudo apt install direnv
```

Lyndeno:
This output confuses me:
```console
[...]

PR #450, opened 2023-12-20
set bash 4.4 as minimum

Mic92:


Mic92:
fixes https://github.com/nix-community/nix-direnv/issues/448

PR #454, opened 2023-12-21
fix devshell

Mic92:


Mic92:
@mergify queue

Mic92:
the bash4.4 regression was introduced in a different commit. Will fix this in a bit.

PR #455, opened 2023-12-21
make version sort compatible with busybox sort

Mic92:
There are too many cases where people have non-coreutils versions in there PATH.


Mic92:
@mergify queue

PR #457, opened 2024-01-02
fix nix_direnv_version terminating direnv

Mic92:
This function is used in our README for a long time to upgrade nix-direnv. Now it's failing execution instead.


Mic92:
@mergify queue

PR #461, opened 2024-01-05
Restore terminfo from calling shell

### LINKED ISSUE #405: Direnv behaves differently, when flakes have identical result
I've created a flake for my build of `st` terminal, and [originally](https://git.sr.ht/~acid-bong/st/tree/a4e24b4c0e795b54bfb8c13559ae0cc58c75b2cf/item/flake.nix) it was overriding some attributes of nixpkgs' derivation. In that case direnv (with `use flake` in `.envrc`) didn't load the environment consistently:
- sometimes it didn't load the executables: `make`, `pkg-config` and/or `gcc`
- sometimes it didn't load the libraries
- sometimes it didn't set up some variables (such as `CC=gcc`)
- Starship's nix module didn't display fully: either had only `impure`, no text or kept the name of different flakes I maintain (I `cd` in and out my flakes), and sometimes it kept the word `impure` when I exited the troublesome flake dir

It was also noticeable that it tried to run the `preInstall` command defined by nixpkgs' derivation of `st`, and `TERMINFO` variable didn't go well with Zsh (I assume Direnv loads asynchronously) -- it produced `(eval):export:1: terminfo: inconsistent type for assignment` (although tried setting it manually, works normally). Retried entering and exiting in Bash, no such issue appears.

I [rewrote](https://git.sr.ht/~acid-bong/st/tree/full/item/flake.nix) the flake from `overrideAttrs` and basically copypasted the build recipe from nixpkgs, now Direnv loads the environment consistently, doesn't run that `export` command.

This also didn't happen in other flakes, like [dwm](https://git.sr.ht/~acid-bong/dwm) (also `overrideAttrs`-based) and [manually patched Victor Mono](https://git.sr.ht/~acid-bong/victor-mono-nf) (`mkDerivation`-based).

bbenne10:
Should REALLY close #405.

### DIFF
--- a/direnvrc
+++ b/direnvrc
@@ -131,11 +131,14 @@ _nix_export_or_unset() {
 _nix_import_env() {
   local profile_rc=$1
 
-  local old_nix_build_top=${NIX_BUILD_TOP:-__UNSET__}
-  local old_tmp=${TMP:-__UNSET__}
-  local old_tmpdir=${TMPDIR:-__UNSET__}
-  local old_temp=${TEMP:-__UNSET__}
-  local old_tempdir=${TEMPDIR:-__UNSET__}
+  local -A values_to_restore=(
+    ["NIX_BUILD_TOP"]=${NIX_BUILD_TOP:-__UNSET__}
+    ["TMP"]=${TMP:-__UNSET__}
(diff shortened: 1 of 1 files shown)


Mic92:
@mergify queue

PR #478, opened 2024-03-31
Fix local flakes when prefixed with "path:" (again)

tie:
This change fixes regression introduced in commit 4bf5d40290981c867d70f6d3c0d505dd03bc2c0d. In particular, we should not be passing flake URI without schema to the nix flake archive command.

This fixes `use flake path:devel#direnv` in `.envrc` that previously errored at `nix flake archive […] "direnv"` step (that should have been `path:devel`).

### DIFF
--- a/direnvrc
+++ b/direnvrc
@@ -247,8 +247,8 @@ use_flake() {
   fi
 
   flake_expr="${1:-.}"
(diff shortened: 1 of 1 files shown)


Mic92:
@mergify queue

Mic92:
@mergify rebase

PR #479, opened 2024-04-01
Fix local flakes when prefixed with "path:" (again)

Mic92:
This change fixes regression introduced in commit
4bf5d40290981c867d70f6d3c0d505dd03bc2c0d. In particular, we should not be passing flake URI without schema to the nix flake archive command.

@mergify queue


Mic92:
https://github.com/nix-community/nix-direnv/pull/478

PR #485, opened 2024-04-14
[README] Show the `hide_env_diff` configuration in `home-manager` section

MathisWellmann:
### Background:
After spending too much time trying to figure out how to silence the large block of text showing the exported variables upon entering the directory I came across an elegant solution described in this thread comment:
https://github.com/direnv/direnv/issues/68#issuecomment-2054033048 

### Quote:
Since direnv 2.34.0, to remove the "export" message completely, add hide_env_diff = true to ~/.config/direnv/direnv.toml in the global section:

```toml
[global]
# https://direnv.net/man/direnv.toml.1.html
hide_env_diff = true
```

### Changes:
Adds the required configuration to `home-manager` section in the `README`.

### Impact:
Will help out everyone (who views the `home-manager` section) to quickly find the required configuration setting.
This will greatly increase user retention as without this, the block of logs produced is a major turnoff IMO.

Thanks for this nice project.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,6 +49,12 @@ In `$HOME/.config/nixpkgs/home.nix` add
       enable = true;
       enableBashIntegration = true; # see note on other shells below
       nix-direnv.enable = true;
+      config = {
+        global = {
+          # Hides the rather large block of text that is usually printed when entering the environment.
+          hide_env_diff = true; 
+        };
+      };
     };
 
     bash.enable = true; # see note on other shells below


bbenne10:
This is a direnv configuration option that we don't *need* to interact with. I would prefer to remain somewhat neutral on how users interact with direnv outside of setting up nix-direnv properly.

This stance is because I'm not entirely sure that I agree with the line that their community has drawn around what is "configuration" - and therefore needs to be in `direnv.toml` and what is "code" - and thus needs to be in `.enrvc` or `.direnvrc`. This PR blurs the lines just a *bit* further, I think. 

I am a soft "no" on merging this. While not technically wrong, this documentation does not belong in our README. Furthermore, this is a relatively bleeding-edge feature (released in direnv 2.34, which is not even in a stable NixOS channel yet) and I don't feel like debugging that with users opening issues against nix-direnv with a too-old direnv version. I spend enough time debugging problems that are direnv or nixos configuration related without this one (probably) making it worse. 

I'll let @Mic92 make the final call here though.

bbenne10:
I have pushed a change to the README in #487 that notes this change in a way I feel is more appropriate for this project. Let me know if that works for you.

woile:
I still have no idea how to configure direnv. I just want to set `hide_env_diff` in the nixos configuration. 
If someone would be so kind to explain me, I'd appreciate. I read #487 but I don't understand it.

bbenne10:
You configure direnv independent of nix-direnv. We cannot control anything relating to the user's direnv configuration as nix-direnv is invoked BY direnv so any changes we make would be too late to be meaningful.

So you configure direnv however direnv should be configured for your platform. On NixOS, you add that setting to `programs.direnv.direnvrcExtra`. With home-manager, you set it in `programs.direnv.config`. And with neither, you put it I. `~/.config/direnv/direnv.toml` directly. 

This mess (that is not made by us or can be cleaned up by us) is why we don't discuss this level of setting in this repo. It is simply not a thing we can meaningfully clarify as it relates directly to our plugin.

PR #487, opened 2024-04-19
Issue a warning if we are falling back to the old shell environment

### LINKED ISSUE #486: Fallback to previous devshell upon failure is unclear
Since https://github.com/nix-community/nix-direnv/pull/420 (I believe), if loading the Nix devshell fails, the previous cached one is used. While I do not necessarily disagree with this behavior, I think that it should be much more clear that this is occuring, as currently one might see errors but think “well it loaded anyway, so those errors may not matter”, especially coupled with the fact that Nix *does* have non-fatal errors. Or the user might just miss those errors in a sea of other output or because they are in the flow. Either way, the shell ends up non-obviously in the “stale” state.

It may also make sense for it to be configurable whether to do this (and knowable from the outside, e.g. to condition on fallback having occurred in the `.envrc`).

bbenne10:
This should resolve #486.

### DIFF
--- a/direnvrc
+++ b/direnvrc
@@ -298,6 +298,7 @@ use_flake() {
       local tmp_profile_rc
       local tmp_profile="${layout_dir}/flake-tmp-profile.$$"
       if tmp_profile_rc=$(_nix print-dev-env --profile "$tmp_profile" "$@"); then
+        # If we've gotten here, the user's current devShell is valid and we should cache it
         _nix_clean_old_gcroots "$layout_dir"
 
         # We need to update our cache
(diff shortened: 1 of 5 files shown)


direnvrc
```diff
@@ -326,6 +326,7 @@ use_flake() {
         # which is probably more useful than nothing.
         # Fallback to use that (which means just leaving profile_rc alone!)
         _nix_direnv_warning "Evaluating current devShell failed. Falling back to previous environment!"
+        export NIX_DIRENV_DID_FALLBACK=1
```
Mic92, line 329: Who is the consumer of this environment variable?

bbenne10:
This is expected to be exported to the calling shell for possible user consumption

-------- Original Message --------

Mic92:
Ok. Do you have a concrete use case for it? If so it would be nice to put in the Readme.

bbenne10:
README updated to note the environment variable. Additionally, I've changed the note about `nix-direnv-watch-file` (which is no longer accurate - oops!) and added a note about `hide_env_diff` in (what I believe to be) a more appropriate way than #485.

Mic92 (approved):

Mic92:
@mergify queue

PR #502, opened 2024-06-16
Codespace glowing fortnight 5957p799jjxhpvww

mustafacco7:


bbenne10:
Same question as your other PR. What is the intent here?

bbenne10:
I see that this adds a devcontainer and a vscode configuration. I personally see no need for the vscode configand I am not convinced that we need a devcontainer at all. Can you explain your thoughts? Without justification, I will close this PR.

PR #511, opened 2024-08-15
replace nixpkgs-fmt with nixfmt-rfc-style

Mic92:


bbenne10:
LGTM. Not sure of the necessity to change to nixfmt-rfc-style, but I don't have strong feelings one way or the other - so long as the style is automated and enforced.

PR #513, opened 2024-08-21
Use ambient Nix when available, with a two stage fallback

### LINKED ISSUE #451: nix-direnv should not bundle nix
Re: https://github.com/NixOS/nixpkgs/issues/270741

The nix-direnv derivation wraps `direnvrc` by prepending the file with `NIX_BIN_PREFIX=${pkgs.nix}`.

As a result:
- On non-NixOS systems it's impossible to use nix-direnv with the system version of nix. This is very surprising and causes a number of issues like warnings for unsupported Nix options.
- On NixOS it doesn't take into account `config.nix.package`, instead opting for the hardcoded `pkgs.nix`

In my opinion it doesn't make sense to bundle nix with nix-direnv. I think it's reasonable to expect nix to be in `PATH` on any system where people would actually want to use nix-direnv.

K900:
First, we try to use the ambient Nix version.
Then, we try to use $NIX_DIRENV_FALLBACK_NIX, which is set by default, but can also be overridden by the user.
Only then, if neither is available, we fail.

Fixes #451.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -2,8 +2,8 @@
   resholve,
   lib,
   coreutils,
-  direnv,
   nix,
+  writeText,
 }:
 
 # resholve does not yet support `finalAttrs` call pattern hence `rec`
@@ -17,11 +17,6 @@ resholve.mkDerivation rec {
     name = pname;
   };
 
-  # skip min version checks which are redundant when built with nix
-  postPatch = ''
-    sed -i 1iNIX_DIRENV_SKIP_VERSION_CHECK=1 direnvrc
-  '';
-
   installPhase = ''
     install -m400 -D direnvrc $out/share/${pname}/direnvrc
   '';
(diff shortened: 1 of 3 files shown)


bbenne10 (approved):
This looks great to me. I appreciate the effort. 

I will wait for @Mic92 to have a chance to review before I merge (though they may get around to it first).

default.nix
```diff
@@ -48,15 +40,26 @@ resholve.mkDerivation rec {
...
+        "$NIX_DIRENV_FALLBACK_NIX" = true;
       };
-      execer = [
-        "cannot:${direnv}/bin/direnv"
-        "cannot:${nix}/bin/nix"
-      ];
+      prologue =
+        (writeText "prologue.sh" ''
+          NIX_DIRENV_SKIP_VERSION_CHECK=1
+          NIX_DIRENV_FALLBACK_NIX=''${NIX_DIRENV_FALLBACK_NIX:-${lib.getExe nix}}
```
Mic92, line 61: Can you mention this environment variable in the README? Otherwise it looks good to me.

K900: Done.

Mic92:
@mergify queue

K900:
Is it just broken now...?

Mic92:
> Is it just broken now...?

"required checks" in mergify.yml needs an update.

PR #514, opened 2024-08-22
ci: also run tests on master

Mic92:


Mic92:
Nothing that can be tested in a pull request really.

PR #521, opened 2024-09-16
also test macos

Mic92:


Mic92:
We could also move tests to nix if we use toybox and bare derivations to allow building nix-shells in the nix sandbox.

Mic92:
Looks like macOS is still broken, but again. those error messages are very hard to read.

bbenne10:
I just looked at these failures and I think this is a weird difference in behavior that I don't fully understand right now. It's telling us that the direnvrc inside the testenv is not approved, but it absolutely should be, since the function I call immediately prior to this (`write_envrc`) immediately approves the contents after writing the envrc...

bbenne10:
Updated to latest master to see if the changes made in the `disallow_fallback` branch help resolve the problems we're seeing here.

PR #536, opened 2024-11-26
Always set nix-direnv-reload as executable

nicknovitski:
I am not sure why, but on docker bind mounts using virtiofs and gcsFuse, `-x` is always true for a file which exists.  This means that, before this commit was merged, when using nix-direnv in manual mode on a devcontainer with docker desktop on mac, `nix-direnv-reload` would be created and then not set as executable, and if they tried to run it users would see an error like `bad interpreter: Permission denied`.  After this commit is merged, these scenarios work.

Here's some discussion: https://github.com/docker/for-mac/issues/5029 

Other tools have run into this: https://github.com/pre-commit/identify/pull/189

### DIFF
--- a/direnvrc
+++ b/direnvrc
@@ -140,9 +140,7 @@ touch "$PWD/.envrc"
 touch -r "$PWD/.envrc" "${layout_dir}"/*.rc
 EOF
 
-  if [[ ! -x "${layout_dir}/bin/nix-direnv-reload" ]]; then
-    chmod +x "${layout_dir}/bin/nix-direnv-reload"
-  fi
+  chmod +x "${layout_dir}/bin/nix-direnv-reload"
 
   PATH_add "${layout_dir}/bin"
 }


direnvrc
```diff
@@ -140,7 +140,7 @@ touch "$PWD/.envrc"
 touch -r "$PWD/.envrc" "${layout_dir}"/*.rc
 EOF
 
-  if [[ ! -x "${layout_dir}/bin/nix-direnv-reload" ]]; then
+  if [[ "$(stat --format=%A "${layout_dir}/bin/nix-direnv-reload")" != *x ]]; then
     chmod +x "${layout_dir}/bin/nix-direnv-reload"
   fi
```
Mic92, line 145: If we are already calling binaries we might as well can just call `chmod` unconditionally. However I am bit suspicious about the error. "Bad interpreter" seems to actually refer `/usr/bin/env bash` not working than this script not having the executable bit.

Mic92: Ah. I re-read your issue again. `access()` is indeed a different syscall from stat.

Mic92, line 145: That fix seems more appropriate:

```suggestion
  chmod +x "${layout_dir}/bin/nix-direnv-reload"
```

bbenne10:
Calling `chmod +x` is idempotent, as far as I understand it. I think that is the better solution.

PR #545, opened 2025-01-16
don't warn about dirty trees

Mic92:
It's to be expected in a git repository that our environment will contain non-committed code. No need to warn about that.


bbenne10 (approved):
Haven't yet tested, but this lgtm.

PR #549, opened 2025-02-14
rely on function autocalling

### LINKED ISSUE #285: error: attempt to call something which is not a function but a set
I use flakes with https://github.com/edolstra/flake-compat to get the good things of flakes while still having backwards compatiblity.

Thus I'd love to `use nix` instead of `use flake`, so I get broader compatibility.

My `shell.nix` looks like this:

```nix
(
  import
  (
    let
      lock = builtins.fromJSON (builtins.readFile ./flake.lock);
    in
      fetchTarball {
        url = "https://github.com/edolstra/flake-compat/archive/${lock.nodes.flake-compat.locked.rev}.tar.gz";
        sha256 = lock.nodes.flake-compat.locked.narHash;
      }
  )
  {src = ./.;}
)
.shellNix
```

When installing nix-direnv, I get this error with `direnv allow`:

```
➤ direnv allow
direnv: loading ~/mydevel/copier/.envrc
direnv: using nix -A devShells."x86_64-linux".default
warning: Git tree '/var/home/yajo/mydevel/copier' is dirty
error: attempt to call something which is not a function but a set

       at «string»:1:2:

            1| (import ./shell.nix {}).devShells."x86_64-linux".default
             |  ^
direnv: nix-direnv: renewed cache
direnv: export ~XDG_DATA_DIRS
```

It seems that it's trying to call `shell.nix` with an empty set `{}`, but `shell.nix` is not a function.

CobaltCause:
Instead of the current behavior of always trying to call the expression as a function, or alternatively, trying to implement autocalling manually with `builtins.isFunction`.

Fixes #285.

### DIFF
--- a/direnvrc
+++ b/direnvrc
@@ -492,11 +492,9 @@ use_nix() {
       if [[ -n $packages ]]; then
         extra_args+=("--expr" "with import <nixpkgs> {}; mkShell { buildInputs = [ $packages ]; }")
       else
-        # figure out what attribute we should build
-        if [[ -z $attribute ]]; then
-          extra_args+=("--file" "$nixfile")
-        else
-          extra_args+=("--expr" "(import ${nixfile} {}).${attribute}")
+        extra_args+=("--file" "$nixfile")
+        if [[ -n $attribute ]]; then
+          extra_args+=("$attribute")
         fi
       fi
 


CobaltCause:
FWIW, it looks like this chunk of code was originally added in 3cca1afdec33bc2b860794718c7e2db0c94c168c and d5a5df7eabcb709fe15a4516301771adb87ef057, the former added the `--file`/`--expr` conditional selection while expecting the expression to *never* be a function in the `--expr` case, the latter of which changed it to expect it to *always* be a function in the `--expr` case. Meanwhile, the `--file` case has been doing autocalling the whole time. I'm not sure why the `--expr` case exists in the first place though, because the `--file path attr` functionality has existed since at least 2.4.

Mic92:
@mergify queue

PR #553, opened 2025-03-16
refactor: Extract flake profile loading into a function

urbas:
Motivation for this change is to make the use_flake function a bit smaller and easier to understand. It should help with future contributions.

### DIFF
--- a/direnvrc
+++ b/direnvrc
@@ -218,6 +218,49 @@ _nix_clean_old_gcroots() {
   rm -f "$layout_dir"/{nix,flake}-profile*
 }
 
+_update_flake_profile() {
+  local layout_dir=$1
+  local profile=$2
+  local profile_rc=$3
+  local flake_uri=$4
+  shift 4
+
+  local flake_inputs="${layout_dir}/flake-inputs/"
+  local tmp_profile_rc
+  local tmp_profile="${layout_dir}/flake-tmp-profile.$$"
+  if tmp_profile_rc=$(_nix print-dev-env --profile "$tmp_profile" "$@"); then
+    # If we've gotten here, the user's current devShell is valid and we should cache it
+    _nix_clean_old_gcroots "$layout_dir"
+
+    # We need to update our cache
+    echo "$tmp_profile_rc" >"$profile_rc"
+    _nix_add_gcroot "$tmp_profile" "$profile"
+    rm -f "$tmp_profile" "$tmp_profile"*
+
+    # also add garbage collection root for source
+    local flake_input_paths
+    mkdir -p "$flake_inputs"
+    flake_input_paths=$(_nix flake archive \
+      --json --no-write-lock-file \
+      -- "$flake_uri")
+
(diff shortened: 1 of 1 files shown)


bbenne10:
I'm not pushing back on this change at all, but I am curious: do you have a *concrete* motivation for this changeset? Is there something that this cleanup does for you immediately? 

I ask because this feels very much like a changeset that will come before another bigger change and I want to stay on top of those. 

Thank you for your contribution and I will merge once I can get this tested locally (obviously assuming no issues).

urbas:
> I'm not pushing back on this change at all, but I am curious: do you have a _concrete_ motivation for this changeset? Is there something that this cleanup does for you immediately?

I was exploring how to add a configurable caching mechanism to nix-direnv. While I was trying to understand the code and figure out where I could drop the cache logic, it occurred to me that the function I extracted is precisely the bit around which caching would fit.

One of my changes was to extract the function to keep the code manageable. The caching logic I'm investigating is not ready for proposal (let alone review), but I thought the extracted function seemed like a valuable and small enough cleanup to contribute on its own.  

> 
> I ask because this feels very much like a changeset that will come before another bigger change and I want to stay on top of those.

The "caching" change I'm investigating should hopefully be very small. Something like this:

```bash
_restore_layout_dir_from_cache() {
  # If a caching command is provided by the user and the restore from cache succeeds (i.e. it reproduces the `.direnv` directory as it should be for the given inputs), then return 0
  # otherwise return 1
}

_store_layout_dir_into_cache() {
  # if a caching command is provided by the user, call it so that it caches the `.direnv` for the given inputs.
}

# This is inside the `use_flake` function
if ! _restore_layout_dir_from_cache "$layout_dir" "$flake_uri" watches; then
  _update_flake_profile "$layout_dir" "$profile" "$profile_rc" "$flake_uri" "$@"
  if [[ "${NIX_DIRENV_DID_FALLBACK:-0}" != "1" ]]; then
    _store_layout_dir_into_cache "$layout_dir" "$flake_uri" watches
  fi
fi
```

> 
> Thank you for your contribution and I will merge once I can get this tested locally (obviously assuming no issues).

bbenne10:
the idea of pluggable cache locations and frameworks seems like it'd be a *bear* to debug (in addition to the current misconfiguration bugs we encounter as the majority of the issues we triage). I think the idea warrants more discussion in an issue and I may hold off on merging this until such a time as that discussion is resolved.

urbas:
@bbenne10: Here's a working prototype of persistent caching: https://github.com/urbas/nix-direnv/pull/2 (I am now using this regularly on my machines.)

Please don't feel like this implementation is what I'm aiming for. I am pretty sure that an upstreamable implementation will look very different.

I also understand that "persistent caching" is a heavy feature, so please feel free to let me know if you think the idea of persistent caching is not suitable for nix-direnv at all.

bbenne10:
I'll discuss those changes at the PR in question. Thank you for an example.

PR #562, opened 2025-05-01
Make the example usage commands less intrusive for non-nix users

mprasil:
This is to humbly suggest making the usage examples less intrusive for non-Nix folks. I frequently use just plain `use flake` in my `.envrc` as I start private repository, because it works fine for _me_, but I usually forget to update it to something less intrusive when I share code with other people.

This leads to a bit of frustration as there are people out there that do use direnv, but don't use Nix. For them the configuration is annoyingly broken. Hitting this snag does not provide the best context to learn about Nix, so I would suggest to perhaps provide usage examples that fall back more gracefully.

This changes some of the examples to make the usage of direnv less intrusive for non-nix folks. I've kept the one-liner nature of these. It's not as nice code, but it's nicer behavior IMO.

### DIFF
--- a/README.md
+++ b/README.md
@@ -159,7 +159,7 @@ pkgs.mkShell {
 Then add the line `use nix` to your envrc:
 
 ```console
-$ echo "use nix" >> .envrc
+$ echo "if has nix; then use nix; fi" >> .envrc
 $ direnv allow
 ```
 
@@ -199,7 +199,7 @@ $ nix flake new -t github:nix-community/nix-direnv <desired output path>
 ### Integrating with a existing flake
 
 ```console
-$ echo "use flake" >> .envrc && direnv allow
+$ echo "if has nix; then use flake; fi" >> .envrc && direnv allow
 ```
 
 The `use flake` line also takes an additional arbitrary flake parameter, so you


bbenne10:
Sigh.

I don't know how to feel about this. This seems like protecting users from themselves. What should happen if direnv's built-in `layout python` is used without python installed? I would expect it to fail loudly (and I expect nix-direnv to fail similarly loudly). Furthermore, users should have approved the envrc after reviewing it, rather than blindly running `direnv allow`.

We do some protection of our users at work when using direnv (by checking for a late enough version of nix-direnv before we call `source_url`, for instance), so there is precedence in our usage of this at my company, but I dont know if I want to make it "mandatory" (because including it here makes it official and therefore the way people will do it more often than not)

mprasil:
Thanks for your input @bbenne10. My motivation here is that I've just seen this complaint multiple times now and I feel like it creates this bad blood among folks that would otherwise be intrigued to try Nix, they just encounter it in context of "this is broken".

As for `direnv allow` or `direnv disallow`, the .direnv I see in the wild often contains couple env variables that are handy and then `use nix` and people that don't use nix and would typically install whatever dependencies there might be some other way then lose the convenience of having project-specific variables set.

And I do think Nix kind of stands out there, because unlike `layout python` where most likely the project requires Python anyways, it's not always the case the project requires Nix. It's just handy tool people use.

But I hear your concern with making the pattern official. Would there be some appetite to perhaps add another short section under usage with some tips on how to make the `.envrc` usage more inclusive? I've seen some interesting snippets even within Nix community like `use flake || use nix` that people might consider as a good starting point. Just to give users some idea that `.envrc` with just single `use nix` in it is not the only way to use this and in fact might not be the most optimal for their specific use case. I'm happy to create PR for that.

(I'm also okay to just let this go, "no" is a valid answer for me as well.)

Mic92:
If someone wants to use it like in their project, they can, but don't think we need to mention this possibility in the README explicitly.

PR #564, opened 2025-05-01
Add details for `nix profile` installation method

synalice:
Previous version of these instructions did not make it very clear that you need to have nix-direnv _together_ with regular direnv.

As a newcomer, I had to learn this the hard way and spend some time debugging this. I'm sure this will be very helpful to other people like me.

Even better, why not even specify direnv as a `buildInput` for `nix-direnv` in `flake.nix`?

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,10 +22,15 @@ CPU load).
 
 ## Installation
 
-> **Heads up**: nix-direnv requires a modern Bash. MacOS ships with bash 3.2
-> from 2007. As a work-around we suggest that macOS users install `direnv` via
-> Nix or Homebrew. There are different ways to install nix-direnv, pick your
-> favourite:
+> [!WARNING]\
+> We assume that [direnv](https://direnv.net/) is installed properly because
+> nix-direnv IS NOT a replacement for regular direnv _(only some of its
+> functionality)_.
+
+> [!NOTE]\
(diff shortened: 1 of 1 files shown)


bbenne10:
Honestly, I am not here to hand hold someone through the direnv's installation steps. I would prefer a global statement above the installation steps (or merely those affected) that states that we assume that direnv is installed properly prior to running these instructions. I'm not sure what I think about merging this, as it doesn't clarify all of the relevant places we assume direnv is installed (the `source_url` instructions are similarly burdened, as an example)

bbenne10:
The failure [here](https://buildbot.nix-community.org/#/builders/1560/builds/147) is the only thing keeping this from merge. Thank you for your help.

synalice:
Uh oh. From what I understand, CI wants README to look like this
```
> [!NOTE] nix-direnv requires a modern Bash. MacOS ships with bash 3.2 from
``` 
instead of this
```
> [!NOTE]
> nix-direnv requires a modern Bash. MacOS ships with bash 3.2 from
``` 

But first snippet is not a valid format for GitHub's [markdown alert](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#alerts)...

I'll try adding two spaces, maybe it would please the CI.

bbenne10:
Thanks for futzing with this to get it to pass existing formatters. I really appreciate the work.

PR #565, opened 2025-05-01
Fix dollar sign usage before commands without showing output in README

synalice:
The dollar signs are unnecessary in this situation, and should not be included. It is easier to copy/paste and less noisy if the dollar signs are omitted when they are not needed.

Detailed explanation:
<https://github.com/DavidAnson/markdownlint/blob/v0.37.4/doc/md014.md>

### DIFF
--- a/README.md
+++ b/README.md
@@ -158,9 +158,9 @@ pkgs.mkShell {
 
 Then add the line `use nix` to your envrc:
 
-```console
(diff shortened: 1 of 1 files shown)


bbenne10 (approved):
I honestly dont have strong feeling about this lint in the slightest, but it does not increase the maintenance burden at all to remove these characters

PR #583, opened 2025-05-27
Drop nix version checks

Mic92:


direnvrc
```diff
@@ -6,7 +6,6 @@ NIX_DIRENV_VERSION=3.0.7
 # min required versions
 BASH_MIN_VERSION=4.4
```
Mic92, line 7: Probably we will have the same issue with the version comparison here at some point. We should investigate the behavior of sort.

bbenne10:
This deserves a minor version bump to get us working again with nix unstable, I think :/

Issue #582, opened 2025-05-26
Nix version detection broken with latest unstable


JM445:
The nix version detection seems to be broken when using latest nix unstable branch.
direnv allow is printing:

`direnv: nix-direnv: minimum required nix version is 2.4 (installed: )
direnv: nix-direnv: Nix version in PATH is too old, wanted 2.4+, got nix (Nix) 2.30pre20250521_76a4d4c2, will attempt fallback`

As far as I know, Nix 2.30 is higher than nix 2.4

bbenne10:
This is a problem in our version detection regex. Frankly, I'm not sure that the nix version check is required any longer. 2.4 is ancient. I'll probably push a change that removes the requirement for nix shortly unless I can think of a better solution that works without affecting our other tool version checking.

PR #593, opened 2025-07-19
Improve version testing logic and add relevant tests

fidgetingbits:
I had run into #582 and before I realized #583 fixed it I already had tweaked the version checks, so figured I'd send a PR anyway since the comments mentioned the versions should be revisited.

This PR modifies `_require_version` to match up to the first non-digit or decimal character, so will drop `pre*` suffix from a version like `2.31pre20250712_b1245123` to become `2.31`, which keeps `sort -c -V` happy. `_required_cmd_version` is also modified to match on a `x.y` or `x.y.z` style version, because with the above nix git version, the `2.31` wouldn't have been matched correctly.

Testing the shell functions with pytest didn't seem worth, so I add a simple bats test file. It may be preferred to build the .bats file in nix so the paths don't need to be passed in environment variables, but I didn't bother for now.


fidgetingbits:
Seems I broke the python tests with the path change to tests/python/, so will take a look.

fidgetingbits:
Ya that's reasonable. It wasn't obvious at the time from a quick poke through the existing pytest approach how to do it easily, so bats seemed like a more natural fit. I agree having 2 testing frameworks is overkill though and half expected it would be frowned upon anyway. I'll look into implementing them in pytest using the PATH trick or similar.

bbenne10:
It might be this evening before I get around to checking more deeply, but bats seems closer to what we need here.

bbenne10:
Still working on this, but it is slow going. Life keeps happening... 

I think we should simply merge and pull out the pytest tests later. @Mic92: any thoughts?

PR #605, opened 2025-08-28
Disallow fallback

bbenne10:
Since there was concern about the default value, I went ahead and made this configurable.

Adds tests and modifies the readme to document the behavior.


bbenne10:
I'll get the tests passing as soon as I can, but that will probably be next week

Mic92:
How do I have to read the new test output? I only see line noise in the test output.

bbenne10:
One of these days I will remember to run `treefmt` before I push...

PR #612, opened 2025-09-17
Fix typo

Wilfred:


bbenne10:
Whoops. Good catch. Thanks for the help :)

PR #627, opened 2025-10-20
Log what causes the cache to be invalidated

### LINKED ISSUE #597: Figure out why a reload happened?
I ran into a scenario where one of my projects re-eval'd the shell environment when entering its directory, but it shouldn't have (none of the watched files changed for example). Is there a good way to see why a given auto-reload would happen? Would a PR be accepted to add opt-in verbose logging for these cases?

winterqt:
Fixes https://github.com/nix-community/nix-direnv/issues/597

### DIFF
--- a/direnvrc
+++ b/direnvrc
@@ -293,21 +293,38 @@ use_flake() {
   local profile_rc="${profile}.rc"
   local flake_inputs="${layout_dir}/flake-inputs/"
 
-  local need_update=0
   local watches
   _nix_direnv_watches watches
+
+  local profile_missing=0
+  if [[ ! -e $profile ]]; then
+    _nix_direnv_info "cache invalidated: profile ($profile) does not exist"
+    profile_missing=1
+  fi
+
+  local profile_rc_missing=0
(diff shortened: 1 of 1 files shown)


bbenne10:
I use Lix as well, so that's not likely the cause. I'll see if I can find a few minutes to poke around (posting from a non-nix enabled system at all right now, so not quite available yet). I'm not sure how I feel about these being unconditionally verbose, but I also don't really know if it is worth it to make them optional.

bbenne10:
I didn't, but getting it merged seemed more important than chasing my tail on that. I'll have to find some time to get it looked over more closely.

llakala:
> Any significant change we make here makes someone upset. We simply cannot win. I believe this is a perfectly fine default, but welcome a PR to disable the functionality, much like manual reloading does. I will not be changing this myself in the near future due to personal reasons affecting my ability to contribute in my (already limited) free time. Thanks for your understanding.
> […](#)

Done in https://github.com/nix-community/nix-direnv/pull/746.

PR #647, opened 2025-12-11
fix: chmod permission error when nix-direnv-reload owned by different user

tennox:
## Summary

Fix `chmod: Operation not permitted` error when `nix-direnv-reload` file is owned by a
different user (e.g., a colleague entered direnv first).
```console
$ direnv allow
direnv: loading /etc/nixos/.envrc
direnv: using flake . --impure --accept-flake-config
chmod: changing permissions of '/etc/nixos/.direnv/bin/nix-direnv-reload': Operation not permitted
```

## Problem

`chmod` cannot change permissions on a file owned by another user.

## Solution

Use `install -m 755 /dev/stdin` instead of separate `cat` + `chmod`. The `install` command
unlinks and recreates the file rather than modifying it in place, which works even when the
existing file is owned by another user.

### Test
Add this to `.envrc`:
```bash
source_url "https://raw.githubusercontent.com/tennox/nix-direnv/fix/chmod-permission-error/direnvrc" \
    "sha256-rXKNtMDtfdKIeU3FnIDmEAyF5NZ3of+Yz2YuU8pYeJI="
```

### DIFF
--- a/direnvrc
+++ b/direnvrc
@@ -96,7 +96,10 @@ _nix_direnv_preflight() {
   # Because direnv_layout_dir is user controlled,
   # we can't assume to be able to reverse it to get the source dir
   # So there's little to be done about this.
-  cat >"${layout_dir}/bin/nix-direnv-reload" <<-EOF
+  # Use install -m instead of cat + chmod to atomically create the file with
+  # correct permissions. This works even when an existing file is owned by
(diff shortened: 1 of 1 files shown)


bbenne10:
Has this fix been tested on Darwin? I worry about its options not matching a nixos or Linux provided binaries.

PR #688, opened 2026-02-08
Release 3.1.1

Mic92:
Release 3.1.1 of nix-direnv


bbenne10:
This version never got released as an artifact. Unsure if we need to do anything about that now, but we should figure out if it was simple oversight or if something bigger happened, I think.

mdaniels5757:
It was released, it just wasn't marked as "latest". I've taken the liberty of doing so.

PR #692, opened 2026-02-09
Avoid `realpath -e` as the flag isn't supported on macOS

ipetkov:


hdgarrood:
fwiw I'm on macos and `resholved` handles this. In my `~/.nix-profile/share/nix-direnv/direnvrc`, I have
```
  if path=$(/nix/store/nixxlz2dfdwmy6r8da5sas4nrnj7sq3z-coreutils-9.10/bin/realpath -e "$(_nix eval --impure --expr "<nixpkgs>" 2>/dev/null)"); then
```
(and this works fine)

bbenne10:
It resolves one part, but not the other. We still haven't gotten to the bottom of why the tests are failing on Darwin, but it does pull in the GNU version of realpath. 

The problem is if you're only using nix-darwin incidentally (i.e. from a project rather than from your home-manager or nix-darwin configuration). We want to be able to continue to support direnv's `source_url`, which means we need to work on "naked" Darwin (unfortunately)

bbenne10:
I am going to get this merged today. The problem has been on my end. I am using this to remind me to test macOS CI runs again, but I simply haven't had the time or motivation. I'll rebase away the last commit and push when I can sit down at a computer.

ipetkov:
@bbenne10 went ahead and rebased and dropped the last commit!

PR #722, opened 2026-04-22
use_nix: watch for the custom `$nixfile` as well

acid-bong:
Originally, `use nix custom.nix` didn't add `custom.nix` to the watchlist. This change adds support for both files and directories as `$nixfile` (which actually was reflected in [README](https://github.com/nix-community/nix-direnv#tracked-files)).

Partially reverts #410

### DIFF
--- a/direnvrc
+++ b/direnvrc
@@ -483,7 +483,15 @@ use_nix() {
     esac
   done
 
-  watch_file "$HOME/.direnvrc" "$HOME/.config/direnv/direnvrc" "shell.nix" "default.nix"
+  local files_to_watch
+  files_to_watch=("$HOME/.direnvrc" "$HOME/.config/direnv/direnvrc" "shell.nix" "default.nix")
+  if [ -f "$nixfile" ]; then
+    files_to_watch+=("$nixfile")
+  elif [ -d "$nixfile" ]; then
+    files_to_watch+=("$nixfile/default.nix")
+  fi
+
+  watch_file files_to_watch
 
   local watches
   _nix_direnv_watches watches


direnvrc
```diff
@@ -483,7 +483,15 @@ use_nix() {
     esac
   done
 
-  watch_file "$HOME/.direnvrc" "$HOME/.config/direnv/direnvrc" "shell.nix" "default.nix"
+  local files_to_watch
+  files_to_watch=("$HOME/.direnvrc" "$HOME/.config/direnv/direnvrc" "shell.nix" "default.nix")
+  if [ -f "$nixfile" ]; then
+    files_to_watch+=("$nixfile")
+  elif [ -d "$nixfile" ]; then
+    files_to_watch+=("$nixfile/default.nix")
```
acid-bong, line 491: i initially wanted to `watch_dir $nixfile`, but that would add unwanted overload, if there's too much files in it

bbenne10: Do you have metrics to back up slowness? I've had relatively good success watching every nix file in nixpkgs, so I am inclined to say "just use `watch_dir`" but I don't want to say that when evidence contradicts me.

bbenne10 (approved):
This looks reasonable.

direnvrc
acid-bong: not really metrics, i only assume that

also, since in Nix `dir` means `dir/default.nix`, and both `use_flake` and `use_nix` only watch the individual files, not directories, so i'll leave watching directories entirely up to the user

acid-bong: just tried with my music directory (i have a Nix shell there with FFmpeg and some tools):
- plain `use nix`: loads instantly
- `watch_dir E-40; use nix` (my copy of E-40 discography weighs 5.3 GB): 3 seconds on first load, 1.5 seconds on repeated visits (but only when entering `~/music`, repeated actions inside it don't trigger anything)

the overhead seems to be really small, but idk. i'll still keep `watch_file` only for consistency

bbenne10: Thanks for the investigation and that seems pretty bad from my perspective (especially since none of the files themselves is actually relevant from a nix perspective).

PR #734, opened 2026-05-11
Print at most 5 newer watched files

### LINKED ISSUE #733: Limit printing number of changed files
I recently updated to a version containing https://github.com/nix-community/nix-direnv/pull/627, and while I do appreciate the extra info, I use nix-direnv in a project that has a few hundred watched `.nix` files, which makes the output a bit verbose.

Would you consider a patch which limited the logging to 5 changed files - perhaps like this?

```
direnv: nix-direnv: cache invalidated: files newer than cache:
nix-direnv: file1.nix
file2.nix
file3.nix
file4.nix
file5.nix
(and 36 others)
```

hdgarrood:
Fixes #733. Example output from testing this locally with `use nix`:

```
> touch flake.nix
direnv: loading ~/Code/mwb/.envrc
direnv: using nix ./shell.nix -A default
direnv: nix-direnv: cache invalidated: profile (/Users/harry/Code/mwb/.direnv/nix-profile-26.05-klzydr6a4fwr1vbg-80675a43cf44443a60e0262cf0ff88cca193456c) does not exist
direnv: nix-direnv: cache invalidated: profile_rc (/Users/harry/Code/mwb/.direnv/nix-profile-26.05-klzydr6a4fwr1vbg-80675a43cf44443a60e0262cf0ff88cca193456c.rc) does not exist
direnv: nix-direnv: cache invalidated: 989 files newer than cache, including:
nix-direnv: /Users/harry/Code/mwb/.envrc
/Users/harry/Code/mwb/nix/deps-manual/persistent-typed-db.nix
/Users/harry/Code/mwb/nix/deps-manual/static-ls.nix
/Users/harry/Code/mwb/nix/deps-manual/buck-worker-proto.nix
/Users/harry/Code/mwb/nix/deps-manual/services.nix
direnv: nix-direnv: cache does not exist. use "nix-direnv-reload" to create it
```

### DIFF
--- a/direnvrc
+++ b/direnvrc
@@ -326,12 +326,12 @@ use_flake() {
     fi
   done
 
-if [[ ${#newer_files[@]} -gt 0 ]]; then
+  if [[ ${#newer_files[@]} -gt 0 ]]; then
     _nix_direnv_info "cache invalidated: ${#newer_files[@]} files newer than cache:"
     echo -n "$_NIX_DIRENV_LOG_PREFIX" >/dev/stderr
     printf "%s\n" "${newer_files[@]:0:5}" >/dev/stderr
-    [[ ${#newer_files[@]} -gt 5 ]] && echo "And $((${#newer_files[@]} - 5)) more" > /dev/stderr
-fi
+    [[ ${#newer_files[@]} -gt 5 ]] && echo "And $((${#newer_files[@]} - 5)) more" >/dev/stderr
+  fi
 
   if [[ $profile_missing -eq 1 || $profile_rc_missing -eq 1 || ${#newer_files[@]} -gt 0 ]]; then
     if [[ $_nix_direnv_manual_reload -eq 1 && -z ${_nix_direnv_force_reload-} ]]; then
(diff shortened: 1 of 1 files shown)


direnvrc
```diff
@@ -318,23 +318,27 @@ use_flake() {
...
 
-  local file_nt_profilerc=0
+  local files_nt_profilerc=0
   local file=
   local newer_files
   for file in "${watches[@]}"; do
     if [[ $file -nt $profile_rc ]]; then
       newer_files+=("$file")
-      file_nt_profilerc=1
+      files_nt_profilerc=$((files_nt_profilerc + 1))
```
bbenne10: You can just do `${#newer_files[@]}` to get the number of entries in a bash array. (This is technically the "length" parameter expansion, but it works for arrays by counting the number of entries). I just tested and this was present in 4.4, which is our target min version.

direnvrc
```diff
@@ -318,23 +318,27 @@ use_flake() {
...
     fi
   done
 
-  if [[ $file_nt_profilerc -eq 1 ]]; then
+  if [[ $files_nt_profilerc -gt 5 ]]; then
+    _nix_direnv_info "cache invalidated: $files_nt_profilerc files newer than cache, including:"
+    echo -n "$_NIX_DIRENV_LOG_PREFIX" >/dev/stderr
+    printf "%s\n" "${newer_files[@]:0:5}" >/dev/stderr
```
bbenne10, line 332: If we rework the logging we can shrink the code paths to just one check, since the `:0:5` syntax works if there are less than 5 elements:

```console
$ x=("a" "b" "c")
$ printf "%s\n" "${x[@]:0:4}"
a
b
c
```

Something like:

```bash
if [[ ${#newer_files[@]} -gt 5 ]]; then
    _nix_direnv_info "cache invalidated: ${#newer_files[@]} files newer than cache:"
    echo -n "$_NIX_DIRENV_LOG_PREFIX" >/dev/stderr
    printf "%s\n" "${newer_files[@]:0:5}" >/dev/stderr
    [[ ${#newer_files[@]} -gt 5 ]] && echo "And $((${#newer_files[@]} - 5)) more" > /dev/stderr
fi
```

bbenne10: (admittedly this is two checks, but it doesn't repeat the printing logic in both exceptional cases)

hdgarrood: That seems fine to me, as long as we are okay with the minor grammar issue of "1 files newer than cache"?

bbenne10: "file(s)" is probably better, but I'm not sure I care enough to conditionally append the "s"...

hdgarrood: I suppose we could actually remove the number from the first line entirely, since we're going to print the number in the "and N more" part anyway

hdgarrood: Do you have a preference between that and "file(s)"?

bbenne10: Let's just remove the number entirely. Apologies for the delay in getting back to you! Work needed my full brain :P

hdgarrood: oh no worries at all! In my mind a turnaround for OSS that's less than one day is pretty good lol

PR #745, opened 2026-05-26
Avoid touching envrc to trigger reloads

lheckemann:


bbenne10:
This looks like a reasonable change, but I'd like to understand why you made it. Can you please explain what motivated this change?

lheckemann:
Currently, when using `nix-direnv-manual-reload`, running `nix-direnv-reload` will touch .envrc and cause the cache to be considered stale right after refreshing it. By watching the profile cache file instead, we get more accurate information on whether the cache is fresh and can reload it automatically if, for example, `nix-direnv-reload` was run in a different terminal.

PR #746, opened 2026-05-27
Allow disabling logging of the reason for cache invalidation

llakala:
I'm using `nix-direnv-manual-reload`, but the logging messages still print, even when I haven't manually reloaded. I also personally find them a bit spammy. Now they can be disabled by setting `NIX_DIRENV_LOG_REASON=1`. I've tested this and can confirm it functions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -343,6 +343,9 @@ nix-direnv also respects the following environment variables for configuration.
   set manually. Leave unset or empty to fail immediately when a Nix
   implementation can't be found on `PATH`.
 
+- `NIX_DIRENV_LOG_REASON`: On reload, print the files that were changed, causing
+  the cache to be invalidated. Enabled by default - set to `1` to disable.
+
 ## General direnv tips
 
 - [Changing where direnv stores its cache][cache_location]
--- a/direnvrc
+++ b/direnvrc
@@ -32,6 +32,8 @@ _nix_direnv_error() { log_error "${_NIX_DIRENV_LOG_PREFIX}$*"; }
 
 _nix_direnv_nix=""
 
+_nix_direnv_log_reason=""
+
 _nix() {
   ${_nix_direnv_nix} --no-warn-dirty --extra-experimental-features "nix-command flakes" "$@"
 }
@@ -73,6 +75,10 @@ _nix_direnv_preflight() {
       ! _require_cmd_version "$direnv" "$DIRENV_MIN_VERSION"; then
       return 1
     fi
+  fi
+
+  if [[ -n ${NIX_DIRENV_LOG_REASON:-} ]]; then
+    _nix_direnv_log_reason="${NIX_DIRENV_LOG_REASON}"
   fi
 
   if command -v nix >/dev/null 2>&1; then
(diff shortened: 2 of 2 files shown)


bbenne10:
The established pattern here is to have a function that is called, which sets a variable internally which can then be checked before taking action. This violates this latter by using a user set environment variable instead. This should be made to work in line with other toggles (see `nix-direnv-manual-reload` for an example)

Additionally, if I am reading this properly, the README documents this variable as one we SET not one we READ. 

I will have another look to ensure I am it reading this improperly in the morning (It is 12:30 am and I am on my phone. I'm a bit fallible at the moment :P)

llakala:
> This violates this latter by using a user set environment variable instead. 

So does the existing `NIX_DIRENV_FALLBACK_NIX` - but sure, I can add a command.

bbenne10:
Let me have a look this evening and I'll respond better then. Apologiea.

>

bbenne10:
`NIX_DIRENV_FALLBACK_NIX` is of a different *kind*. It isn't a *feature toggle* but rather a customization applied to packaged versions of direnv that allow specifying a wrapped `nix` to invoke directly.

`NIX_DIRENV_FALLBACK_NIX` is *largely* a packager tweak that is applied in nixpkgs to make it work appropriately when packaged in nixpkgs.

There is an argument to be made about making changing `NIX_DIRENV_FALLBACK_NIX` to make it consistent with the manual reloading behavior as well, but this falls somewhat short as this variable is meant to be set in nixpkg wrappers rather than in user-facing `.envrc`. However, intention does not matter in practice and I recognize that somewhere someone is probably doing something unintended with `NIX_DIRENV_FALLBACK_NIX` and I can't (and don't care to) stop them from this workflow.

However,what I *can* do is say that we should be consistent in our interfaces *moving forward*. I was explicit in #627 when I said this interface should mimic the interface exposed by manual reloading. This does not do that. 

Frankly, as the one that'll be maintaining the code you are asking to merge, I think this is a reasonable ask in order to minimize user confusion, maintainer confusion and the effort required to integrate new code. Thanks for understanding.

llakala:
Yeah, of course! Was just mentioning it offhand. Happy to.

bbenne10:
I had to go back and look at `NIX_DIRENV_FALLBACK_NIX` and this is 50% me regurgitating that understanding and 50% me clarifying (to both you and me) how I want to move forward into the future. Re-reading this, I think it's harsher (maybe "colder"?) than I intended. Sorry about that.

thunze:
Is there any reason (parts of) these logs don't go through `log_status`? If they did, they would respect `DIRENV_LOG_FORMAT` and could be silenced via `DIRENV_LOG_FORMAT=""`.

bbenne10:
I think that's a combination of simple oversight on our (read: my) part and a desire to print some logs to stderr.

thunze:
FWIW, `log_status` seems to log to stderr as well.

bbenne10:
I have opened #754 to get our logging in line with what direnv wants. 
It changes us to use `log_status` everywhere and I've changed our stance such that any "internal" logs that don't disappear by setting direnv's log_filter to "^$" is now a test failure. Once that MR is merged, I think that anything you want to do with regard to eliminating logging from nix-direnv should now go through `log_filter` in direnv and I may begin to undo any logging toggles added to the codebase. If you want additional functionality, I am of the opinion that direnv needs better knobs to tune this (One regex is not quite enough, but I don't have ownership of that codebase to fix the shortcomings I see in how that is handled there.)

If that MR gets merged, the right path forward is to set "log_filter" to something that ignores the `nix-direnv:` prefix. I'm not entirely sure how I would go about writing this, as go regex's have neither negative lookahead nor negative lookbehind - which makes this regex MISERABLE to write - but I believe that to be a problem in direnv that could be fixed by changing the regex engine in use.

I'm not sure yet what I want to do with the work on this branch.

bbenne10:
Overtaken by #754. If there is specific reason to want or need something from this branch, please open a new ticket with the specific motivation.

Issue #749, opened 2026-05-30
Doesn't work in ro directories


dpc:
I'm doing some sandboxing/isolation tooling, and I've just hit an issue where even when the environment did not change at all, `nix-direnv` fails when the directory is RO due to:

https://github.com/nix-community/nix-direnv/blob/8e58e00a8d71f773f1d389414c57f48ef15c33a7/direnvrc#L100

Is this some hard requirement with reasons I missed or an implementation detail that could be improved? I'm thinking about avoiding re-writting that file if identical already to what it should be.

bbenne10:
Yes - the reason is that we must ensure that the reload script file is owned by your user to set the right permissions. You cannot `chmod` a file not owned by you, so we recreate the file on each run to ensure the file is owned by your user and not another user colliding with the layout directory. 

Honestly though, the answer in both your case and the colliding user case is the same: Move your layout directory. Direnv gives you the tool to solve this globally, so we should probably use it.

[The direnv wiki has an entry for this.](https://github.com/direnv/direnv/wiki/Customizing-cache-location) I think you should use it, as it sidesteps the problem entirely by allowing you to decide where to put your cache. I cannot predict everything and - as stated - your PR invalidates another use case.

dpc:
> Yes - the reason is that we must ensure that the reload script file is owned by your user to set the right permissions.

What if the file is already owned by the same user? I mean ... 99.999% of the time it will be - there is no other user to begin with. Could we skip it then? That's what I mean in "I'm thinking about avoiding re-writting that file if identical already to what it should be."


> Honestly though, the answer in both your case and the colliding user case is the same: Move your layout directory. Direnv gives you the tool to solve this globally, so we should probably use it.

It would work for me, but then my tool collides with the defaults of any other `direnv` user out there, so I have to keep telling them to change it.

My sandboxing tooling puts a ro bind over the working copy only in certain situations, temporarily. So the direnv setting likely didn't change.

bbenne10:
I don't think we will ever get this to a reasonable state that accounts for every scenario. What if their default chmod mask is different from mine? What if their fs doesn't allow symlinks? What about bash on windows? There are a million different scenarios that I cannot account for and a read only direnv-enabled directory feels as likely to me as just about any of them. It fels more prescient for you because you're living it and I appreciate that, but I am trying to balance my interest and availability in maintaining this software with your totally valid but also quite niche usage scenario.

[...]

PR #750, opened 2026-05-30
Avoid rewriting unchanged reload helper

dpc:
For now just trying to self-help myself with #749 .

Disclosure: this change is made by an LLM, and it looks OK to me, but I don't really know that much about nix-direnv myself, so :shrug: .

### DIFF
--- a/direnvrc
+++ b/direnvrc
@@ -56,6 +56,54 @@ _require_cmd_version() {
   _require_version "$cmd" "${BASH_REMATCH[1]}" "$required"
 }
 
+_nix_direnv_reload_script() {
+  cat <<-EOF
+#!/usr/bin/env bash
+set -e
+if [[ ! -d "$PWD" ]]; then
+  echo "Cannot find source directory; Did you move it?"
+  echo "(Looking for "$PWD")"
+  echo 'Cannot force reload with this script - use "direnv reload" manually and then try again'
+  exit 1
+fi
+
+# rebuild the cache forcefully
+_nix_direnv_force_reload=1 direnv exec "$PWD" true
+EOF
+}
+
+_nix_direnv_install_reload() {
+  local layout_dir=$1 reload tmp desired existing
+  reload="${layout_dir}/bin/nix-direnv-reload"
+  desired=$(_nix_direnv_reload_script)
+
+  if [[ -O $reload && -x $reload ]]; then
+    existing=$(<"$reload")
+    if [[ $existing == "$desired" ]]; then
+      return 0
+    fi
+  fi
+
+  tmp=$(mktemp "${layout_dir}/bin/.nix-direnv-reload.XXXXXX") || return
+
+  # Write to a temporary file first so the existing helper stays usable if
+  # writing the new script or marking it executable fails.
+  if ! printf "%s\n" "$desired" >"$tmp"; then
+    rm -f "$tmp"
+    return 1
+  fi
+
+  if ! chmod +x "$tmp"; then
+    rm -f "$tmp"
+    return 1
+  fi
+
+  if ! mv -f "$tmp" "$reload"; then
+    rm -f "$tmp"
+    return 1
+  fi
+}
+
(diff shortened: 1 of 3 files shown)


dpc:
Works perfectly so far.

bbenne10:
TIL about "test -O". Please add a test where the profile is owned by a different user and I will merge when everything is passing. Thanks.

direnvrc
```diff
@@ -56,6 +56,45 @@ _require_cmd_version() {
...
+_nix_direnv_force_reload=1 direnv exec "$PWD" true
+EOF
+}
+
+_nix_direnv_install_reload() {
+  local layout_dir=$1 reload tmp desired existing
+  reload="${layout_dir}/bin/nix-direnv-reload"
+  desired=$(_nix_direnv_reload_script)
+
+  if [[ -f $reload ]]; then
```
bbenne10: This whole if block feels like it could boil down to this, right?

```bash
if [[ -O "$reload" && -x "$reload" ]]; then
  existing=$(<"$reload")
  if [[ "$existing" == "$desired" ]]; then
    return 0
  fi

tmp=$(mktemp "${layout_dir}/bin/.nix-direnv-reload.XXXXXX") || return
if ! printf "%s\n" "$desired" >"$tmp" || ! chmod +x "$tmp" || ! mv -f "$tmp" "$reload"; then
  rm -f "$tmp"
  return 1
fi
```

`test -O` will check the file's existence and I'm not sure that `chmod +x` without creating the file is helpful.

dpc:
> TIL about "test -O".

Same. :D

tests/test_reload_helper.bats
```diff
@@ -0,0 +1,36 @@
...
+  local layout_dir="$TESTDIR/layout"
+  local reload="$layout_dir/bin/nix-direnv-reload"
+  mkdir -p "$layout_dir/bin"
+  ln -s "$NIX_DIRENV_RELOAD_HELPER_OTHER_OWNER" "$reload"
+
+  run bash -c 'PWD=/nix/store; source "$DIRENVRC"; _nix_direnv_install_reload "$1"' -- "$layout_dir"
+  assert_success
+  assert [ -f "$reload" ]
+  assert [ ! -L "$reload" ]
+  assert [ -O "$reload" ]
```
bbenne10:
```suggestion
assert [[ -f "$reload" && -O "$reload" ]]
```

There's no benefit to avoiding the bash test extensions here. And we can drop `! -L` as we have already proven it is a regular file and so it is definitely not a symlink.

direnvrc
```diff
@@ -56,6 +56,41 @@ _require_cmd_version() {
...
+
+  if [[ -O $reload && -x $reload ]]; then
+    existing=$(<"$reload")
+    if [[ $existing == "$desired" ]]; then
+      return 0
+    fi
+  fi
+
+  tmp=$(mktemp "${layout_dir}/bin/.nix-direnv-reload.XXXXXX") || return
+  if ! printf "%s\n" "$desired" >"$tmp" || ! chmod +x "$tmp" || ! mv -f "$tmp" "$reload"; then
```
bbenne10: What is this pattern? Why are we not just writing to the desired file and then making it executable with `chmod`?

Is the intention to make the write+chmod atomic as a single unit?

bbenne10 (commented):
I think this is the last round. Sorry for not thinking about this the first time through.

direnvrc
dpc: `mv` is making switching atomic, and impossible to be stuck with a half-written file e.g. in case of disk drive being full

dpc:
> I think this is the last round. Sorry for not thinking about this the first time through.

Not a problem, I'm happy to follow up more. Thank you for taking time to look at it. :)

dpc:
@bbenne10 Just doing Sunday evening PR grooming.

bbenne10 (approved):

PR #754, opened 2026-06-03
Fix logging

bbenne10:
This fixes our logging to work well with the builtin direnv logging tooling and adds tests to ensure we don't revert on this in the future. I have also retooled some of the other testing bits in light of a better understanding of the direnv logging primitives.

The only disruptive changes here may be the redirection of stderr for some additional nix invocations that were printing some extraneous lines to stderr during testing and the fact that I've removed the deprecated `nix_direnv_watch_file`. It's been spitting out deprecation notices since late 2023. I think that's long enough. We should still bump the major version to indicate backwards breaking changes for that. Happy to rebase that commit away if desired.


bbenne10:
Oops. Forgot shellcheck/treefmt was going to get invoked on tests too. I'll resolve those formatting issues this evening.

bbenne10:
@Mic92: do you want to review? Or do you trust me to merge without? I've been leaving my PRs open for your review, but I don't want to force your involvement. Just let me know what you want to do.

PR #790, opened 2026-08-20
Don't refresh gcroots

### LINKED ISSUE #786: Auto-reload causes issues with multiple terminals open
With the changes in #745 direnv now watches `.direnv/flake-profile-...` for changes. This seems like a great idea but it causes issues when the same `.envrc`/direnv is being used in multiple terminals.

It's really a combination of two PR's causing issues:

* In #745 a new watch is added, [`$profile_rc`](https://github.com/nix-community/nix-direnv/pull/745/changes#diff-ad5f6477d1306d9e37dcc8e4ac962fa6534ea673533df51e7940eda7e5841748R386). 
* In #631 this is touched during the [`_nix_refresh_gcroots`](https://github.com/nix-community/nix-direnv/pull/631/changes#diff-ad5f6477d1306d9e37dcc8e4ac962fa6534ea673533df51e7940eda7e5841748R366). 

Because `$profile_rc` is being watched, whenver there is an mtime mismatch it will re-load the direnv. But, because the `_nix_refresh_gcroots` is run during the reload, this touch and modify the mtime of the `$profile_rc`.

To reproduce:

1. Open a terminal that will be used for this demonstration (terminal 1)
2. Clone https://github.com/KaiSforza/nix-direnv-repro
3. run `direnv allow`. This will fetch the `./direnvrc` from #745.
4. Wait for the direnv to finish loading...
5. Open a new terminal in the same directory (terminal 2)
6. In terminal 1, press enter a few times to see that the caching does work and you don't get a reload in that terminal. Wait for it to load.
7. In terminal 2, press enter. You can verify the normal caching works by pressing enter more than once. It should load normally. Wait for it to load.
8. Again in terminal 1, press enter. It will reload here.
9. Go back to terminal 2 and press enter. It will reload again.
10. (repeat as many times as you want)

If terminal 1 and 2 are both reloaded at the same time, that is the only way that this will not happen, since the `.rc` file will be touched at about the same second, which is the resolution of direnv's `$DIRENV_WATCHES`.

bbenne10:
This logic (introduced in #631) conflicts with #745. Because the motivation for #631 was nh's cleanup functionality and that functionality now has a blessed & upstreamed augmentation for this sort of behavior (namely `nh clean --keep-one`) we can safely revert this changeset without reducing general utility.

Closes #786

### DIFF
--- a/README.md
+++ b/README.md
@@ -343,6 +343,23 @@ nix-direnv also respects the following environment variables for configuration.
   set manually. Leave unset or empty to fail immediately when a Nix
   implementation can't be found on `PATH`.
 
+#### Notes on nh integration
+
+`nh` is a modern nix helper utility that generally makes interacting with the
+nix tooling much better and more friendly. In the past, nix-direnv had some
+specific functionality to update gcroots whenever the environment was refreshed
+so that `nh` would consider those gcroots "fresh" and would not clean them up
+when `nh clean` was invoked.
+
+Modern versions of `nh` (4.4.0+) introduce a `--keep-one` argument to the
+`clean` subcommand, which specifically keeps one revision of every direnv's
+gcroots, which allows easily keeping the latest revision of every nix-direnv
+related gcroot so you're not redownloading the state unnecessarily.
+
+Because of this functionality (and because it interfered with other common
+usage), we have reverted our changes supporting this functionality. Please use
+`--keep-one` if you want a specific gcroot keps, regardless of age.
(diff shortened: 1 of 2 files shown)


bbenne10:
I'll add a note about `--keep-one` in the README sometime this weekend and then merge if there's no other feedback

bbenne10:
...and I immediately noticed a typo. New MR incoming...

